Issue #268, opened 2026-07-26
T3 content tier fails on a fresh checkout: data/generated/audio.lua is never generated


johnjohto:
**One line summary:** After following the Developer Setup guide (ROM import via `scripts/setup.sh` / `setup.ps1`), the T3 content tier fails because three suites need `data/generated/audio.lua` — and nothing in the documented flow generates it.

**How to reproduce:**
1. Fresh clone, on any OS
2. `scripts/setup.sh --rom /path/to/red.gb` (runs `tools/build_data.py`)
3. `luajit tests/run_tests.lua`

**What happens:**
```
FAIL cries rendered for the full dex (0)
FAIL mod_registry_tests: tests/mod_registry_tests.lua:473: module 'data.generated.audio' not found
FAIL parity_intro: tests/parity_intro.lua:126: attempt to index field 'audio' (a nil value)
```

**Why:** `tools/build_data.py` has no audio step (its `--only` list doesn't include audio). `tools/extract/audio.py` can render the WAVs and write `data/generated/audio.lua`, but nothing calls it — no CLI entry point, no import from `build_rom_data.py`, no script. The in-app importer goes the Lua route (`assets/generated/audio/programs.bin`) and never writes `audio.lua` either. CI can't catch this because it has no ROM and skips T3 entirely. My guess: it passes on the maintainer's machine because of a `data/generated/audio.lua` left over from an older flow.

**What should happen:** either the dev build generates `audio.lua` (wire `extract/audio.py` into `build_data.py`), or the three suites skip cleanly when it's absent. Happy to send a PR for whichever direction you prefer — just say which.

**Anything else:** found while setting up a Windows dev environment. Separate from this, T3 also has a handful of Windows-only shell portability failures (`/dev/null`, ffi `setenv`, `mkdir -p`, `ls` globs in the save-editor tests); I'll file those with a fix.

PR #269, opened 2026-07-26
Make the T3 content tier run on Windows

johnjohto:
Follow-up to #267, one level down: with a ROM imported, the T3 content tier had the same Unix-shell assumptions inside the suites themselves.

## What was broken on Windows

- `tests/run_tests.lua` redirected to `/dev/null` when chaining the tier runners, so every chained tier "failed"
- `tests/mod_runtime_tests.lua` called ffi `setenv`/`unsetenv`, which msvcrt doesn't have (`_putenv` with an empty value unsets)
- `tests/modkit_tests.lua` captured exit codes with POSIX `; echo 0` (cmd needs `/v:on` and `!errorlevel!`), called `python3` (it's `python` on Windows), and used `mkdir -p` / `rm -rf` — cmd's `mkdir` creates parents by itself and was creating a literal `-p` directory in the repo root
- the save-editor suites globbed `save.lua.bak-*` backups with `ls`
- `tools/save-editor/Catalog.lua` scraped mod flags with `ls` — this one affects the real save editor on Windows, not just the tests

## Verification (Windows 11, Git Bash, LuaJIT 2.1, Red ROM imported)

`luajit tests/run_tests.lua`: 12 failures → 3. The remaining 3 all trace to `data/generated/audio.lua` never being generated, which fails the same way on a fresh Linux or macOS checkout — that's #268 and needs a maintainer decision, so it's deliberately not touched here.

Unix commands are unchanged behind `package.config` checks; CI is the check that Linux still agrees.

Note: this branch builds on #267 (it reuses the `tests/fs_io.lua` helpers), so it should land after that one — or I can rebase if you'd rather take them independently.


bryanthaboi:
@johnjohto merge conflicts

johnjohto:
Merge conflicts resolved. I merged the latest dev in. The conflicts were in tests/fs_io.lua and tests/run_modkit.lua: I kept FsIo.globPrefix since run_save_editor_tests.lua still uses it, and took your example_* exclusion in run_modkit. ROM-free tiers and the save-editor suite are all green locally.

PR #387, opened 2026-07-29
Add native Android second-screen companion

kennethreitz:
## TL;DR

- Adds an OLED-black live companion on the physically smaller Android display while the game remains on the larger one.
- Shows location, party HP/EXP, status, badges, cash, and the current battle opponent without mutating game state.
- Leaves single-screen devices unchanged and keeps all focus, touch, and controller input with the game.

## What changed

Android builds now publish a small presentation-only JSON snapshot from the live game every 300 ms when its contents change. The snapshot includes:

- current location and map coordinates
- the six-party roster with HP, status, level, and EXP to the next level
- earned badges and cash
- the current battle opponent

`GameActivity` discovers active displays each time it resumes. When the game is running on the largest active display and a distinct smaller display is available, it shows a native OLED-black `Presentation` on that smaller display. Display IDs are never persisted, so panel power changes and Android display reassignment are handled on the next resume.

The presentation is both non-focusable and non-touchable. It is owned by the game activity rather than launched as a second activity; this keeps controller focus with the game and avoids Android's resumed-activity ANR behavior for windows that intentionally cannot take focus.

If only one display is active, the game was launched on the smaller display, or display presentation fails, the companion is withheld and normal behavior continues.

## Why

Dual-screen Android handhelds such as the AYN Thor have enough room to surface useful RPG state without covering the game. The companion makes that space useful while remaining optional by construction and preserving the ordinary single-screen path.

## Validation

- `lua tests/second_screen_test.lua` — 15/15 checks passed
- `scripts/build_android.sh` — `assembleEmbedNoRecordDebug` succeeded
- Live AYN Thor validation:
  - game on the 1920×1080 upper display
  - companion on the 1240×1080 lower display
  - live party/location data rendered correctly
  - game remained the focused activity
  - presentation window reported `NOT_FOCUSABLE` and `NOT_TOUCHABLE`
  - no ANR after sustained dual-display operation

### DIFF
--- a/mobile/android/love/src/main/java/org/love2d/android/CompanionPresentation.java
+++ b/mobile/android/love/src/main/java/org/love2d/android/CompanionPresentation.java
@@ -1,0 +1,333 @@
+package org.love2d.android;
+
+import android.app.Presentation;
+import android.content.Context;
+import android.graphics.Canvas;
+import android.graphics.Color;
+import android.graphics.Paint;
+import android.graphics.RectF;
+import android.graphics.Typeface;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.view.Display;
+import android.view.View;
+import android.view.Window;
+import android.view.WindowManager;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.io.BufferedInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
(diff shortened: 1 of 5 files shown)


bryanthaboi:
screenshot please

kennethreitz:
image

i can provide a real screenshot later

bryanthaboi:
can u do me a favor? only additions or modifications to the api to make this possible as a mod should be committed and then release as a mod. if this is possible without modifications at all, release as a standalone mod.

PR #464, opened 2026-07-30
iOS: stop fusing mods; Pokewalker moves to its own repo

mresnick67:
Small follow-up to #452, from dogfooding the merged build on a real iPhone.

## Un-fuse mods from iOS `game.love`

A mod fused into the app bundle is read-only: the mod manager's **Delete can't remove it** (it reappears every launch), and an imported newer version can never fully replace it. iOS builds now pack no mods — mods install as `.zip`s at runtime (launcher → MODS → Import mod .zip), the same install/upgrade/delete lifecycle as every other platform. `build_android.sh` never fused anything, so this just makes iOS match.

## Pokéwalker moves out of the tree

With nothing fusing it, the mod has no reason to live in this repo, and the in-tree copy was already stale (0.1.0 here vs 0.2.0 at its canonical home, [Gen1ReComp-Pokewalker](https://github.com/mresnick67/Gen1ReComp-Pokewalker) — 0.2.0 adds move learning via the rare-candy flow and a properly paged credit dialog). This PR removes `mods/pokewalker/` and points `docs/ios-install.md` at the mod repo's releases instead.

Bookkeeping that follows: the `event:save.loaded` coverage-debt entry returns to `gate_meta_coverage` — its coverage came from the mod's test suite and leaves with it (the gate fails in both directions, so this is the honest state).

If you'd rather keep the mod in-tree as a gallery entry, happy to go the other way and bump it to 0.2.0 here instead — say the word.

## Testing

Full local reproduction of the `ci` workflow on this branch against current `dev`: all tiers pass, mod lint clean, fingerprint gate 16/16, meta-coverage gate 167/167 (with the restored debt entry). On-device: verified the delete/upgrade lifecycle now works on iPhone with the mod imported as a zip.


mresnick67:
Apologies here; my previous PR (https://github.com/bryanthaboi/gen1recomp/pull/452) vendored my Pokewalker mod directly with the iOS fixes. I am making a small change here to remove the mod from iOS build, so a user can optionally install it seperately.

PR #489, opened 2026-07-30
Android: hardware step-counter, enabling Pokewalker

mresnick67:
## What this is

The Android half of the step-sync seam from #452: `love.system.syncHealthSteps()`
now exists on Android too, so the same step-sync mods (e.g.
[Pokéwalker](https://github.com/mresnick67/Gen1ReComp-Pokewalker), which #464
moved to its own repo) work on official Android builds without a patched APK.
Same file contract as the iOS `GRHealthBridge`; the mod is byte-identical across
platforms.

Nothing in the base game calls the new seam — with no consumer mod installed, the
only change to the shipped APK is one dormant manifest permission.

## How it works

Same JNI route as the SAF picker (`wrap_System.cpp` → `System.cpp` →
`common/android.cpp` → `GameActivity`):

- **`GameActivity.syncHealthSteps`** — one-shot read of the hardware
  `TYPE_STEP_COUNTER` sensor: cumulative since boot, counted by the OS whether or
  not any app is running — no Google Fit / Health Connect dependency. The reading
  is anchored in `SharedPreferences` so a walk is never credited twice; a reading
  *below* the anchor means the phone rebooted, which re-anchors without crediting
  (steps between reboot and next sync are lost — errs on the honest side). Deltas
  (50k clamp per sync, same as iOS) merge into `steps_pending.json` in the save
  identity dir — merged with any unconsumed delivery, never overwritten, written
  atomically via tmp+rename.
- **`ACTIVITY_RECOGNITION`** declared in the app manifest. Android 10+ gates the
  sensor behind this runtime permission: the first `syncHealthSteps()` call shows
  the system prompt, and on grant the read runs immediately
  (`onRequestPermissionsResult`); Android 9 and older prompt nothing. The build
  script's permission trim (RECORD_AUDIO / WRITE_EXTERNAL_STORAGE) leaves it
  alone. Sensor-less devices (some tablets) log and return without writing.
- Devices without the sensor, denied permission, or a `null` activity all fail
  soft — the Lua side just never sees a pending file.

## Build script fix (independent, same area)

`build_android.sh` now shadow-builds from a space-free temp dir when the checkout
path contains spaces — ndk-build is GNU make underneath and fails on paths like
`~/xCode Projects/...` ("Your APP_BUILD_SCRIPT points to an unknown file"). The
shadow persists across runs so gradle/NDK builds stay incremental. Checkouts at
space-free paths (including CI) build exactly as before.

## Docs

`mobile/ANDROID.md` gains a step-bridge section (JNI route, anchoring semantics,
permission behavior, reboot caveat).

## Testing

- `assembleEmbedNoRecordDebug` builds clean from this branch on current `dev`
  (JDK 17, SDK 34, NDK 25.2.9519653); the APK declares `ACTIVITY_RECOGNITION`
  and carries `syncHealthSteps` in both `classes.dex` and `liblove.so`.
- Field-tested: an equivalent build shipped as the Pokéwalker project's sideload
  APK; an independent tester confirmed step tracking works on a real phone
  (shake-to-step included).
- Desktop/iOS untouched: the new native code is Android-only; `System.cpp`
  returns `false` on other platforms in this tree, and iOS keeps its own #452
  bridge.

### DIFF
--- a/mobile/ANDROID.md
+++ b/mobile/ANDROID.md
@@ -50,6 +50,29 @@ chosen file under the app save directory as `picked_rom.gb`,
 from that folder on Choose / refocus; see `docs/launcher.md`. The APK payload
 itself remains data-free (no embedded ROM or generated cache).
 
+### Step bridge (Pokéwalker mod)
+
+`love.system.syncHealthSteps()` → `GameActivity.syncHealthSteps` (same
+JNI route as the picker: `common/android.cpp` →
+`modules/system/System.cpp` → `wrap_System.cpp`). The Java side does a
+one-shot read of the hardware `TYPE_STEP_COUNTER` sensor (cumulative
+since boot, counted by the OS whether or not any app runs), anchors the
+reading in `SharedPreferences` so a walk is never credited twice
+(a reading below the anchor means the phone rebooted → re-anchor without
+crediting), and stages the delta as `steps_pending.json` in the save
+identity dir — the same contract as the iOS `GRHealthBridge`. Nothing in
+the base game calls it; the consumer is the
+[Pokéwalker mod](https://github.com/mresnick67/Gen1ReComp-Pokewalker),
+installed as a mod `.zip` at runtime (its SYNC STEPS option defaults
+off).
+
+Android 10+ gates the sensor behind the `ACTIVITY_RECOGNITION` runtime
+permission (declared in `app/src/main/AndroidManifest.xml`; keep it out
+of the build script's permission trim). The first
+`syncHealthSteps()` call shows the system prompt; on grant the sensor
+read runs immediately (`onRequestPermissionsResult`,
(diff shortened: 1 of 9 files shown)


mresnick67:
Don't have an Android to test this personally, but indication (from an earlier build) is that the change worked as intended: https://discord.com/channels/1019387038820216882/1532402597204463817/1532461951077060668

Shared a new build from this PR which is waiting to be tested: https://discord.com/channels/1019387038820216882/1532402597204463817/1532482780997615736

mresnick67:
Tested on a real device, working as intended: https://discord.com/channels/1019387038820216882/1532402597204463817/1532487851453976757.

bryanthaboi:
update ur branch to resolve conflict

mresnick67:
@bryanthaboi  should be fixed now.

PR #516, opened 2026-07-31
Cap low-HP alarm to a fixed cycle in modern ruleset

sirj0k3r:
Faithful ruleset keeps the siren looping for the whole red-HP duration.
Modern ruleset (Gen 6+) now stops it after a fixed cycle (2 trips
through the 2-beep source, ~1.5s) instead of blaring for the rest of
the battle. Gated behind `ruleset.modernLowHealthAlarm`


bryanthaboi:
Make mod instead. If mod requires additional changes to the API, make a PR with just what is necessary for your mod to work.

PR #517, opened 2026-07-31
Gen 6+ exp sharing under modernExpShare

sirj0k3r:
Extracted exp-awarding logic out of `enemyMonFainted` into a shared
`awardExp()` function, so both faints and catches can grant exp.

Under `ruleset.modernExpShare` (modern_clean.lua):
- Participants no longer split exp among each other — each gets the
  full, undivided amount instead of 1/participants.
- Every non-participant, non-fainted party mon gets a flat 50% exp
  share automatically — no Exp Share item needed.
- Catching a wild mon now also grants exp, using the same split logic
  as a faint would.

Gated behind the ruleset flag, off by default — faithful gen1 behavior
(EXP.ALL item, split among participants, no exp on catch) is untouched.

### DIFF
--- a/src/battle/BattleState.lua
+++ b/src/battle/BattleState.lua
@@ -3208,7 +3208,10 @@ function BattleState:onFaint(battler)
   end
 end
 
-function BattleState:enemyMonFainted()
+-- Exp for the defeated enemy, shared by the faint path (enemyMonFainted)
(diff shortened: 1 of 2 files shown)


bryanthaboi:
Make mod instead. If mod requires additional changes to the API, make a PR with just what is necessary for your mod to work.

PR #524, opened 2026-07-31
Claude/multi game low end support  and some fixes o16gz5

kaosregulator:
## What this does

Three self-contained changes to the LÖVE2D engine: a performance tier so
lower-end devices can run smoothly, plus two latent-bug fixes and a static
analysis baseline found while auditing the codebase. No gameplay logic
changes — every fix is either presentation or a crash/error path.

## 1. Graphics performance tier for low-end devices

A new **OPTIONS → PERFORMANCE** setting scales the port's optional,
non-faithful extras down for weaker hardware:

| Tier | 3D TILT | GBC FX | Survey ZOOM | FPS ceiling |
|------|---------|--------|-------------|-------------|
| HIGH | on | on | on | none |
| BALANCED | off | off | on | none |
| LOW | off | off | off | 60 |
| **AUTO** *(default)* | picks a tier from the device | | | |

- `AUTO` reads the device once at boot: ARM handhelds (e.g. the RG34XXSP) →
  LOW, phones/tablets and very-low-core desktops → BALANCED, a normal
  desktop → HIGH. **Every existing `options.lua` and the common desktop
  case resolve to HIGH, so behavior is unchanged.**
- Nothing touches game logic (fixed-step off `dt`), so all tiers play
  identically — only how much eye-candy the renderer does changes.
- Clamps are applied to the *live* state without rewriting stored options,
  so a lower tier hides your TILT/GBC FX/ZOOM/FPS choices and a higher tier
  restores them exactly.
- New `src/core/Performance.lua` (zero requires, like `GameVersion`); wired
  into `Game:applyOptions`; `Zoom.offsetRange` floors survey at FIT on LOW.

## 2. Two latent bugs found by static analysis

- **`music.volume` mod hook crashed the game** (`src/core/Music.lua`):
  `applyVolume` was defined above `local state = {…}`, so its hook-context
  reads bound to the nil global `state`. Any mod registering `music.volume`
  hit *"attempt to index a nil value (global 'state')"*. Forward-declared
  `state`.
- **Link-battle failures showed the wrong message** (`src/link/Tournament.lua`):
  `local battle, why = isHost and newHost() or newGuest()` dropped the
  second return (`why`), so every failure showed the generic "Link battle
  can't start" instead of the specific reason — and when a host's
  `newHost()` returned nil, the `or` wrongly fell through to `newGuest()`.
  Split into explicit `if/else`.

## 3. Static-analysis baseline

Added a tuned `.luacheckrc` + `scripts/lint.sh`. High-signal by design
(keeps undefined globals/locals, unused values, unreachable code; mutes the
cosmetic categories the codebase lives with). `luacheck src` goes from
**185 → 7 warnings, 0 errors**. Also dropped one dead `require`.

## Testing

- New ROM-free regression test `tests/engine/music_volume_hook_state.lua`
  (reproduces the Music crash on the old code).
- Options-row golden updated for the new PERFORMANCE row.
- `scripts/test.sh` — **all ROM-free tiers pass** (engine 35/35, modkit 2/2).

## Docs

`docs/new-features.md`, `README.md`, and `docs/anbernic-rg34xxsp.md` updated
for the performance tier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ __pycache__/
 .*
 !.github/
 !.gitignore
+!.luacheckrc
 
 # Android build outputs / local SDK path / packaged payload (keep love-android sources)
 mobile/android/app/build/
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,43 @@
+-- Static-analysis config for `luacheck` (https://luacheck.readthedocs.io).
+--
+-- Run it over the engine with:  luacheck src   (or scripts/lint.sh)
+--
+-- The point is a high-signal baseline: the categories left on are the ones
+-- that catch real defects -- undefined globals/locals (the class that hid a
+-- `music.volume` crash: applyVolume read a `state` that was still the nil
+-- global), unused values, unreachable code, redefinitions. The cosmetic
+-- categories the codebase deliberately lives with (a `self`/`dt` an
+-- interface requires but a given method ignores, documented empty
+-- fall-through branches, the odd long line) are muted so they don't drown
+-- the signal.
+
+std = "luajit"
+
+-- LÖVE exposes `love` as a mutable table: games assign their callbacks onto
+-- it (love.wheelmoved, love.run, ...), so it is a regular global, not
+-- read-only -- otherwise every callback registration reads as a violation.
+globals = { "love" }
+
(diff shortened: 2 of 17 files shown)


bryanthaboi:
update ur branch, claude

PR #538, opened 2026-07-31
Add battle.low_health_alarm mod hook

sirj0k3r:
Wraps the low-HP siren start/stop in updateFx behind a
battle.low_health_alarm hook. ctx exposes on (mirrors
self.lowHealthAlarmOn) and battle; a mod can reshape the toggle (e.g.
force ctx.on false after some cycle budget) before letting vanilla
act on it.

No-op when unclaimed — vanilla start/stop behavior is unchanged.


bryanthaboi:
both fail a test

Issue #512, opened 2026-07-31
Issue import ROM Yellow on iOS


santi3100:
### One line summary

Issue import ROM Yellow on iOS 

### Which game were you playing

Yellow

### Which build are you running

iOS

### Were any mods on

No

### Version

0.1.45

### Discord username (optional)

santi3100

### Where in the game

import rom

### Screenshot, or a detailed description if you couldn't take one

[

<img width="1406" height="956" alt="Image" src="https://github.com/user-attachments/assets/cf36a380-ef27-419f-b370-4162cf62d0c1" />

](url)

### How to make it happen

none

### What should have happened

none

bryanthaboi:
Closing as a duplicate of #482, same iOS crash on Import ROM, and that thread already has it confirmed by three people including you.

PR #539, opened 2026-07-31
Fix #482: guard love.system.pickFile so Import ROM cannot crash

### LINKED ISSUE #482: Issue while trying to import ROM on iOS
### One line summary

When clicking on Import ROM, the app crash

Red

Multiple platforms

### Were any mods on

Yes

### Which mods (if any were on)

_No response_

### Version

0.1.44

_No response_

### Where in the game

Nope

Image

### How to make it happen

1. click on import rom (any version)

### What should have happened

The file manager should appear

### Anything else

_No response_

hernan0078:
Fixes #482 (and #512, absorbed into it).

## The crash

From the screenshot in #482:

```
src/import/RomImporter.lua:1112: attempt to call field 'pickFile' (a nil value)

Traceback
[love "callbacks.lua"]:228: in function 'handler'
src/import/RomImporter.lua:1112: in function 'choose'
src/import/RomImporter.lua:2203: in function <src/import/RomImporter.lua:2142>
main.lua:519: in function <main.lua:502>
```

## Why

`love.system.pickFile` is a **native bridge, not part of LÖVE**. It only exists
on builds that compiled one — Android, and iOS builds patched by
`mobile/ios/patch_love_src.py`. On a build without it the field is nil.

`RomImporter` sends iOS down the same path as Android:

```lua
local mobileOS = love.system.getOS()
local android = mobileOS == "Android" or mobileOS == "iOS"
```

…and all three mobile pick sites called the field unguarded. That is why the
reports say **"any version"** — nothing about this is version specific, so Red,
Blue and Yellow all hit the same line.

The part that makes it a shame: **the fallback already exists.** `choose` is
written to handle a device with no document picker —

```lua
self.notice = {
  status = "No picker available, copy your ROM into:",
  detail = love.filesystem.getSaveDirectory(),
}
```

— and `love.system.createFile` at its one call site was already guarded exactly
this way. Those three `pickFile` calls just weren't, so the code written for
this case could never be reached.

## The fix

One small helper that answers `false` when the bridge is absent, used at the
three call sites. A build without a picker now degrades to the
copy-into-the-save-folder flow instead of dying — and on iOS that path works:
`UIFileSharingEnabled` exposes the app's Documents folder in the Files app and
`GRBootstrap` sweeps what lands there into the save directory. So affected
users get a working import route immediately, without waiting for a rebuilt
IPA.

Behaviour on a build that *does* have the bridge is unchanged — covered by a
test.

## Tests

New `tests/rom_importer_no_picker_test.lua` (wired into `run_tests.lua`) covers
Import ROM, Import mod and Import save with the bridge missing, and asserts the
picker is still used when present. 12 checks.

Reverting the one-line-per-site change reproduces the reported error exactly:

```
FAIL Import ROM does not crash without a picker:
     ./src/import/RomImporter.lua:1157: attempt to call field 'pickFile' (a nil value)
```

`scripts/test.sh --quick` on this branch: **ALL TIERS PASSED** (engine 38/38,
mod-SDK 2/2; T3 skipped, no ROM in the checkout).

## Note on the separate question of *why* the bridge is missing

This change makes the app degrade gracefully, which is worth doing regardless.
It does not answer why the shipped iOS build has no `pickFile` — that looks
like the release build not running `mobile/ios/patch_love_src.py`, which is
worth a look on your side, since with the bridge present the picker is the
nicer flow.

One other thing I noticed while in `scripts/build_ios.sh`, not touched here to
keep this PR focused: `pack_game_love` packs `tools/rom_manifest.json` and
`tools/rom_manifest_blue.json` but **not** `tools/rom_manifest_yellow.json`, so
iOS builds ship without the Yellow manifest. Happy to send that as its own PR
if useful.


bryanthaboi:
did u test this on all mainline environments? mac / linux / windows/ android?

hernan0078:
> did u test this on all mainline environments? mac / linux / windows/ android?

No. I only focus on creating an iOS port 

I’m a content creator and I was wondering if I could include an ipa file that already has the mod fixed and working on iOS devices so I can do a video about it .

bryanthaboi:
Can u fix the rest of the crap u mentioned in the PR and I'll happily merge it in tn

hernan0078:
Sorry, can you clarify what do you mean by the “rest of the …”. What would you want me to fix in the PR?

Sent with Shortwave <https://www.shortwave.com?utm_medium=email&utm_content=signature&utm_source=aGVybmFuMDc4QGdtYWlsLmNvbQ==>

On Sat Aug 1, 2026, 01:01 AM GMT, bryanthaboi ***@***.***> wrote:
[image]bryanthaboi left a comment (bryanthaboi/gen1recomp#539) <https://github.com/bryanthaboi/gen1recomp/pull/539#issuecomment-5148736450>

bryanthaboi:
It says something in the description of other stuff that could also make iOS file picking work better or something

bryanthaboi:
Maybe I misread and this'll fix iOS file picker as is?

bryanthaboi:
I'll just merge no worries

hernan0078:
The only issue I found is that the file picker works but it needs a restart of the app to show the ROM was added.

Sent with Shortwave <https://www.shortwave.com?utm_medium=email&utm_content=signature&utm_source=aGVybmFuMDc4QGdtYWlsLmNvbQ==>

PR #543, opened 2026-07-31
Dual screen: render.compose hook + second-screen bridge

BartInTheField:
## Enabling seam: `render.compose` + second-screen bridge

A seam that makes a second screen possible as a *mod capability*, 

Works together with the simple DS presentation mod (world-top / UI-bottom).
https://github.com/BartInTheField/gen1recomp-ds-mod

   | Overworld | Battle |
   |:---:|:---:|
   | image | image |

### What it adds
- **`render.compose` hook** in `Renderer:endFrame`. A mod wrap receives
  `(next, renderer, ctx)`. Return `true` (without calling `next`) to take over
  the whole window; call `next` to run the engine's normal composite and just
  decorate it. 
  `ctx` carries the finished `worldCanvas` + `uiCanvas` with their SGB
  `zones`/`worldZones`, `worldActive`, the frame metrics, `renderer:blitCanvas(...)`
  (a palette-correct canvas blit, lifted from the internal `blit` closure so
  there's one implementation), and the `secondScreen` bridge.
- **`Renderer:blitCanvas(...)`** — the internal composite blit, promoted to a
  public method so a mod can draw either canvas into an arbitrary screen rect
  with correct SGB shading.
- **`SecondScreen.lua` + Android Presentation bridge** (`love_android_secondary_*`
  in `common/android.cpp`, secondary-display plumbing in `GameActivity.java`) —
  the optional transport for driving a real second physical display.

### Explicitly NOT here
This does not add `DualScreen`/`DualBattle`, or battle-layout changes, or dual screen functionality by itself. Just a hook other can use.

### Verification
`scripts/test.sh --quick` → **ALL TIERS PASSED** (T1/T2/T4; T3 needs a ROM).
Ships the three artifacts your meta-gate (`gate_meta_coverage`) requires: a
public-API unit test (`tests/engine/render_compose_seam.lua`), automatic no-mod
parity via `gate_hooks`, and docs (`docs/modding.md`). The Android second-display
path is compile-only here and wants on-device confirmation.

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -187,5 +187,18 @@ composited and before touch controls draw. The window-space viewport contains
 and `dpiY`, so a tool can use the letterbox margins without drawing over the
 playfield or pushing an updating game state.
 
+`render.compose` wraps the whole-window composite in `Renderer:endFrame`. It
+receives `(next, renderer, ctx)`; returning `true` without calling `next` hands
+the mod full control of the window, while calling `next` runs the engine's
+normal single-window composite so the mod can decorate around it. `ctx` carries
+the finished `worldCanvas` and `uiCanvas` with their SGB `zones` / `worldZones`,
+`worldActive`, the frame metrics (`ww`, `wh`, `pw`, `ph`, `ox`, `oy`, `vpw`,
+`vph`, `scale`, `Sx`, `Sy`, `dpiX`, `dpiY`), `renderer:blitCanvas(...)` for a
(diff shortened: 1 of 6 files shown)


BartInTheField:
Putting it in draft, because on Android it is still acting somewhat weird. Working on it

img_0419_1024

BartInTheField:
Much better, actual only needed a fix in the mods code. This API works great

image

PR #546, opened 2026-07-31
feat(mods): pokemon_recolor — a derived palette per species

serivt:
Adds `mods/pokemon_recolor`, a GRAPHICS mod that gives each species its own
colour palette instead of the ten the engine shares across all 151.

Vanilla assigns palettes by class through `icons.byDex` and
`PaletteFX.monPal`: every starter line is one green, twenty-four species
share a single grey, and Charmander and Vileplume are both `REDMON`. This
derives a palette per species from reference artwork (Gen 3) and applies it
to 120 of them, leaving 31 on vanilla where that read better.

**Nothing is replaced.** The Game Boy pic is already the right size, the
right silhouette and already in the player's cache, so the mod ships four
RGB triples per species and repaints the player's own art at install time.
It carries no images at all.

image-1785540854464
image-1785540863230
image-1785540859526

## Route A checklist

| | |
|---|---|
| `modkit validate --base imported` | ok |
| `modkit validate --base imported --strict` | ok |
| `modkit lint` | ok, no ROM-derived content |
| `modkit pack` | ok, 8 files |
| `tests/` asserting the stated effect | 4038 checks |
| `.modkitignore` keeping the suite out of the package | yes |
| `README.md` — one sentence, then the commands | yes |
| `mod.card` — summary, author, tags, differences, credits, compat | yes |
| `CHANGELOG.md` heading matches `manifest.version` | `1.0.0` |
| No ROM-derived bytes | ships no images |
| No bare `error()` / `assert()` in callbacks | `mod.log:*` only |

```
python3 tools/modkit.py validate mods/pokemon_recolor --base imported
python3 tools/modkit.py lint mods/pokemon_recolor
luajit mods/pokemon_recolor/tests/pokemon_recolor_test.lua
```

## How the ramps are derived

`tools/derive_palettes.py`, offline, against the player's own imported
cache. Two matchers, chosen per species after reviewing the whole dex:

- **`histogram`** walks the Game Boy and reference cumulative colour
  distributions in step, so each GB tone lands on the reference colour
  holding the same tonal position by area. Good where the artwork shades
  smoothly. 40 species.
- **`kmeans`** clusters the reference colours by luminance and takes each
  cluster's most-used colour, keeping two strong separated colours apart
  instead of averaging across their boundary. 80 species.

Both are then forced to four visually distinct steps and rescaled onto the
near-white/near-black anchors the hand-tuned GBC palettes use. That last
step is not cosmetic: without it the ramps keep only ~74% of vanilla's
contrast and read washed out at 40px.

The choices live in `ramps.lua` next to their result and are read back
before regenerating, so the file is both the record and the input.
Regeneration is idempotent.

## Two engine details worth flagging

**`trueColor` is required.** Without it the renderer re-shades the repaint
into the active palette's four greys on the way to the screen and the work
is thrown away.

**`DexEntryMenu` does not resolve through `Assets`.** `PartyMenu:199`,
`HallOfFame:57`, `TrainerCard:25` and `OakSpeech:60` all call
`newImage(Assets.resolve(path))`; `DexEntryMenu:44` calls `newImage(path)`,
so the Pokedex shows the untouched cache file while every other screen
shows the repaint. The mod routes around it through the `pokemon.sprite`
hook, which runs inside `Sprites.path`. **This looks like a one-line engine
bug** and would affect any mod deriving battle pics — happy to open a
separate PR for `Assets.resolve(path)` there if wanted.

## Scope

Battle pics only. Tilesets, overworld sprites, party icons, the HUD and the
font are untouched, and the suite asserts those non-effects: no tileset
flagged, no icon registered, and every repainted species keeps its stats,
learnset and types.

## Notes for review

- 31 species are deliberately left on their vanilla palette. They are
  recorded in `ramps.lua` with `variant = 1` and carry no ramp, so nothing
  repaints them and nothing flags them.
- The suite injects a filesystem that hides the imported cache, so it
  exercises the no-cache branch and gives the same result whether or not
  the machine running it has a ROM imported.


bryanthaboi:
mods do not go in this repo, you can post in the discord, in your own repo, and add to the mod index. see the wiki and the #pkmn-links section of the discord

PR #548, opened 2026-08-01
Bundle Dramatic Shape voxel mod as a shipped mod

kaosregulator:
Vendor the DramaticShapeVoxelMod (v1.2.1) into mods/dramatic_shape/ so the 3D voxel diorama ships with the game and is toggleable from the in-game mod manager, rather than being a separate download.

It's a native mod API 2 render-pipeline mod: it registers a `voxel` drawWorld pipeline (extruded terrain, depth-buffered occlusion, voxel character slabs, shadow map) and a `tiltshift` worldPresent post-pass, plus over-the-shoulder 3D battles fought on the map. Purely presentational -- nothing reaches collision, movement, triggers, or scripts.

Bundled the runtime + metadata only: manifest.json, mod.card, README.md, main.lua, lib/ (31 modules), data/ (voxel_heights, battle_arenas). Left out the authoring-only trees the mod's own .modkitignore excludes -- tests/ probes and the Python voxel-carving tools -- and its dev CHANGELOG.

The engine's Manifest.validate accepts it (api 2, game_version "0.0.0-dev || >=0.1.37 <2.0.0", which matches this engine's 0.0.0-dev / modApi 2), all 31 Lua files parse clean, and scripts/test.sh stays green.

Note: bundled mods default to enabled, so 3D-BTL (on-map battles) is on out of the box and hotkeys 3/5 shift from TILT/GBC FX to VOXEL/V-GRID (both still reachable from OPTIONS). The voxel free-roam camera itself still defaults OFF. The mod has no LICENSE of its own yet -- worth adding one before redistributing the packaged game.

Claude-Session: https://claude.ai/code/session_01Q6bFAiQyZ5jDmewsbB4LG9

### DIFF
--- a/mods/dramatic_shape/README.md
+++ b/mods/dramatic_shape/README.md
@@ -1,0 +1,40 @@
+# Dramatic Shape Voxel Mod
+
+A mod for the [Pokémon Gen 1 Recompilation
+Project](https://github.com/bryanthaboi/pokemon-gen1-recomp-project).
+
+The overworld as a 3D diorama. Terrain is extruded into real geometry,
+occlusion comes from a depth buffer rather than a y-sort, characters stand
+as leaning sprite slabs, a shadow map throws real cast shadows across
+whatever they land on, and an optional tilt-shift pass sells the
+miniature-model look.
+
+And battles fought on that world rather than on a white field. When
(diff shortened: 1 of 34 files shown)


bryanthaboi:
I don't think so Claude

Issue #488, opened 2026-07-30
Can't talk to NPC on bench in some Pokémon Centers


DonelBueno:
### One line summary

Read title

### Which game were you playing

Red

### Which build are you running

Windows

### Were any mods on

No

### Version

0.1.44

### Where in the game

Vermilion, Safrom and Fushia Pokémon Centers.

### Screenshot, or a detailed description if you couldn't take one

<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/5390094e-6894-42ae-bde4-2fcaba3a6f3c" />

### How to make it happen

Go to the one of the cities above, enter the pokémon center and try to talk to the NPC on the bench. You can't interact with him.

### What should have happened

You should be able to talk to him.

sebastian926:
This also happens in Saffrom Pokemon Center as well.

DonelBueno:
Same thing in Fushia.

So Vermilion, Saffrom and Fushia so far.

bryanthaboi:
Does he talk in every poke center in the real game?

DonelBueno:
> Does he talk in every poke center in the real game?

Yes, he does. You can interact with every NPC in the real game and they all say something.

PR #549, opened 2026-08-01
Fix bench guys unreachable in some Pokémon Centers (#488)

kaosregulator:
You can't talk to the person on the bench in the Vermilion, Saffron, and Fuchsia Pokémon Centers (Cinnabar too). Nothing happens on A.

Root cause: bench guys are hidden events (data/events/bench_guys.asm). Each seat sits against the x=0 wall, reachable only from (1,y) facing LEFT. The engine gated the interaction on `h.facing` -- the hidden-object trigger arg from hidden_events.asm -- but for these four centers that arg is "up", a direction the seat can never be faced from, so the guy was silent. The direction pokered's PrintBenchGuyText actually checks against the player is the bench_guys.asm facing, which the extractor already stores as `textFacing` ("left" for every bench guy). Gate on that instead.

The other seven centers had facing == textFacing == "left", so they were unaffected either way; this only changes the four whose two directions disagreed.

parity_bench_guys.lua gains the reachability invariant (every labelled bench guy is faced from the left) and a functional check that drives the real tryHiddenObject: facing the seat now triggers a previously-broken bench guy and returns its text, while the stale "up" no longer does. Verified against a freshly built Red data cache.

Claude-Session: https://claude.ai/code/session_01Q6bFAiQyZ5jDmewsbB4LG9

### DIFF
--- a/src/world/OverworldController.lua
+++ b/src/world/OverworldController.lua
@@ -1824,9 +1824,16 @@ function OverworldState:tryHiddenObject(fx, fy)
     end
   end
 
-  -- bench guys (data/events/bench_guys.asm)
+  -- bench guys (data/events/bench_guys.asm).  Gate on textFacing -- the
+  -- bench_guys.asm direction PrintBenchGuyText actually checks against the
+  -- player -- not h.facing, the hidden-object trigger arg.  The two disagree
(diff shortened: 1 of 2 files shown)


DonelBueno:
Already fixed.

PR #622, opened 2026-08-01
Add EUR (Spanish) Red/Blue ROM support

jherediagu:
## What

Adds support for the EUR Spanish ROMs — Pokémon Edición Roja (`fc17c5b9…`) and Edición Azul (`7715e7b1…`) — alongside the US dumps.

A Spanish ROM imports into the same Red/Blue tab as its US counterpart, sharing the same cache and saves. The game plays fully in Spanish, including dialogue, menus, accented fonts, the metric Pokédex (`AL 1,9m / PE 120,0kg`), the `EDICIÓN ROJA/AZUL` title ribbon and credits, translated trade nicknames, and town map names.

## How

The Spanish releases are rebuilds of the US ROMs. Data banks retain their US layout, while the far-text banks and some code-bank tables are shifted.

The shift-matching [[einstein95/pokered-es](https://github.com/einstein95/pokered-es)](https://github.com/einstein95/pokered-es) disassembly builds byte-identical ROMs, so its `.sym` files provide the Spanish addresses for every symbol required by the manifest. This is the same technique used by `make_blue_manifest.py` with `pokeblue.sym`.

* **`tools/make_es_manifest.py`** (new) derives `rom_manifest_red_es.json` and `rom_manifest_blue_es.json` from the shipped US manifests. It re-resolves all ~3269 symbols and inline audio addresses from the Spanish `.sym`, then re-parses only the language-dependent metadata: `charmap`, `fontCharmap`, dynamic text substitutions, presets, credits, trades, town map names, and the title ribbon.

* **`GameVersion` / `RomImporter`**: each version now lists its accepted dumps through `roms = { us, es }`. ROMs are routed by SHA-1, and the cache marker records which dump was imported.

* **Metric Pokédex**: EUR Pokédex entries store height and weight in decimeters and hectograms, using one byte less than the feet/inches representation. Both extractors now support this through `dexUnits` and `dexUnitLabels`. Imperial equivalents are still derived so existing consumers continue to work, while `DexEntryMenu` displays the localized captions.

* **Language-agnostic `tools/extract/` parsers**: Route 23 badges are resolved by label instead of display string; SGB `IF DEF` palette gates respect `ASM_DEFINES`; Blue's intro Pokémon is handled correctly for EUR Blue, where Jigglypuff is inherited from JP Blue; version-gated slots and title-ribbon `INCBIN`s are supported; and Game Corner, town map, and credits sanity checks no longer depend on English text.

Spanish Yellow — Amarillo (`1dc24203…`) — is a clean dump, but there is currently no matching disassembly. It cannot be supported yet, so the launcher now recognizes the hash and explains this instead of implying that the ROM is invalid.

## Validation

* `tools/verify_rom_data.py` exits with code 0 for both Spanish ROMs. Azul uses the new `--define _BLUE` flag. All 16 datasets and 495 generated PNGs extracted from the retail ROMs match the source-backed pipeline byte-for-byte and pixel-for-pixel.

* The runtime Lua extractor decodes all 2585 dialogue labels from the Spanish ROMs with zero unknown bytes.

* `scripts/test.sh --quick` passes.

Regeneration is documented in `docs/extraction-notes.md`:

```bash
make red blue
python3 tools/make_es_manifest.py --pokered-es <checkout>
```

This requires the `pokered-es` repository and RGBDS 1.x.

new commit: 

**Fix EUR title ribbon, custom-name label and manifest packaging**

The 64px EDICION ROJA/AZUL ribbon was drawn with the two-piece layout meant for the 80px US RED VERSION art, reading past the image edge and. cutting the word short; narrower ribbons now draw whole, right-aligned in the VERSION_TILES band the way the ROM loads them.

The naming menu's custom-name row used the engine catalog instead of the extracted names_list.asm entry, so it read NEW NAME on a Spanish import where the game says NUEVO N.

The desktop, Android and RG34XX packaging lists name every manifest explicitly and were missing the two EUR ones, which would have made ROM import fail in a built app while dev stayed green -- the same way the Yellow manifest shipped missing once. iOS derives its list from GameVersion.lua and already picked them up.

**Skip the US-Red content tier on a localized import**

T3 asserts dialogue that is true of the US English release and of nothing else. An EUR import answers those assertions correctly for its own ROM (SEMILLA, ROJO/ASH/JAIME, 34 credit screens), so the tier reported nine failures the code was not responsible for. Detect the imported release from the extracted name presets and skip, leaving the save editor and link tiers running.


bryanthaboi:
This is cool I gotta look more in depth at this

hernan0078:
Nice work — the shift-matching `.sym` trick is neat.

Two things from the iOS side.

I would like to test this on an iPhone, since nobody has. One obstacle first: `patch_love_src.py` raises the deployment target on `love.xcodeproj` only, and `liblove` stays at 8.0, which Xcode 26+ refuses outright — so iOS does not currently build from `dev` at all. I have a one-function fix and will put it up separately; once that is in I can report back on whether Spanish import works on device.

For the Yellow gap: I wrote [gbromdiff](https://github.com/hernan0078/gbromdiff), which diffs two ROMs bank by bank and separates "this bank is translated text" from "this bank is the same code with pointers moved". It does not build anything — just less tedium for whoever attempts `pokeyellow-es`. Yours if it is useful.

bryanthaboi:
@hernan0078 i included gbrom diff in tools!

hernan0078:
This is a great PR — and I think the one gap in it can be closed. I have Spanish Yellow (`1dc242039218fba50928d1afb66b70565b6b9daf`, the same dump this PR lists as unsupported) importing and playing on top of exactly this architecture.

**The premise is right; the conclusion doesn't follow.** There is no shift-matching disassembly for Amarillo and there may never be one. But the addresses don't have to come from a disassembly, because the cartridge states them.

A translated cartridge isn't a rewrite. The Spanish and English Yellow ROMs are **72% byte-identical**, and 32 of their 62 used banks sit at exactly the same offsets — so they align bank by bank. Once aligned, the ROM's own pointers can just be read:

- Gen-1 dialogue is reached through `text_far` (`$17`, address, bank). The `$17` is untouched code, so it relocates by alignment — and the pointer behind it is the cartridge saying where it moved its own text. That recovers ~2,500 dialogue labels whose content is completely different.
- Pokédex descriptions are read out of their own entry blocks (149).
- Map headers come from the pointer table plus its parallel bank table (101 at once), found exactly because we know every entry the English ROM holds.
- Runs of dialogue are walked block by block between two placed labels, accepted only when the walk closes exactly on the label that ends the run.

Result: **all 4,459 manifest symbols plus the 252 inline `{bank, address}` entries**, and `tools/build_rom_data.py` completes every stage — 18 data files, 510 assets, real Spanish (`¡Hola! ¿Puedo ayudarte en algo?`, moves `ÁCIDO`, Pikachu `kind = "RATÓN"`).

Your `dexUnits` work needs no changes: the generator detects the metric layout and emits `dexUnits: "metric"` with `AL`/`PE`. One warning there, since it cost me an hour — the layout has to be decided for the **whole table**, not per entry. TENTACOOL's Spanish entry reads `$09 $C7 $01 $17 $17`, so on its own it looks imperial and comes out 199 inches tall.

Verified against the English manifest as the baseline for what correct looks like:

| | English (truth) | Spanish (derived) |
|---|---|---|
| text addresses just after a terminator | 2,677/2,686 (99.66%) | 2,675/2,686 (99.59%) |
| consecutive labels one text block apart | 2,518 exact | 2,660 exact |
| pairs the block chain can't reconcile | 10 | 10 |
| distinct symbols sharing a destination | 0 | 0 |

**The honest caveat.** Over half the addresses come from reading a relocated `text_far`, and only **44 of those 2,694 sites sit inside a byte-identical run** — the rest are interpolated. A wrapper is five bytes, so an interpolated site can land next door and read a valid pointer to the *neighbouring label's* text: a real address, in range, pointing at real Spanish. Nine labels around `_CantDepositLastMonText` did exactly that, each shifted one block, and no range, order or plausibility check noticed. I found it by reading the decoded Spanish against the English and fixed it (the block walk bridges bad anchors, and its exact-closure requirement is what makes the correction safe). So: every check passes and every spot-check is correct, but I'd call it very good rather than proven. A `pokeyellow-es` would still be the thing that proves it.

Happy to submit this as a follow-up PR on top of #622 — manifest plus the recovery tooling under `tools/es/`, ~1.1 MB for the manifest, same shape as your `_es` ones. Or I can hold it until this lands, or hand it to you to fold in. Whatever you and @bryanthaboi prefer.

One thing I won't include: the extracted Spanish script itself. That's Nintendo's text and belongs in nobody's repo — the manifest carries addresses and the character map only, exactly like the existing ones.

Worth adding that the method isn't Spanish-specific. French and German Yellow disassemblies exist but aren't needed for this route; anyone with a European cartridge can generate its manifest the same way.

jherediagu:
> This is a great PR — and I think the one gap in it can be closed. I have Spanish Yellow (`1dc242039218fba50928d1afb66b70565b6b9daf`, the same dump this PR lists as unsupported) importing and playing on top of exactly this architecture.
> 
> **The premise is right; the conclusion doesn't follow.** There is no shift-matching disassembly for Amarillo and there may never be one. But the addresses don't have to come from a disassembly, because the cartridge states them.
> 
> A translated cartridge isn't a rewrite. The Spanish and English Yellow ROMs are **72% byte-identical**, and 32 of their 62 used banks sit at exactly the same offsets — so they align bank by bank. Once aligned, the ROM's own pointers can just be read:
> 
> * Gen-1 dialogue is reached through `text_far` (`$17`, address, bank). The `$17` is untouched code, so it relocates by alignment — and the pointer behind it is the cartridge saying where it moved its own text. That recovers ~2,500 dialogue labels whose content is completely different.
> * Pokédex descriptions are read out of their own entry blocks (149).
> * Map headers come from the pointer table plus its parallel bank table (101 at once), found exactly because we know every entry the English ROM holds.
> * Runs of dialogue are walked block by block between two placed labels, accepted only when the walk closes exactly on the label that ends the run.
> 
> Result: **all 4,459 manifest symbols plus the 252 inline `{bank, address}` entries**, and `tools/build_rom_data.py` completes every stage — 18 data files, 510 assets, real Spanish (`¡Hola! ¿Puedo ayudarte en algo?`, moves `ÁCIDO`, Pikachu `kind = "RATÓN"`).
> 
> Your `dexUnits` work needs no changes: the generator detects the metric layout and emits `dexUnits: "metric"` with `AL`/`PE`. One warning there, since it cost me an hour — the layout has to be decided for the **whole table**, not per entry. TENTACOOL's Spanish entry reads `$09 $C7 $01 $17 $17`, so on its own it looks imperial and comes out 199 inches tall.
> 
> Verified against the English manifest as the baseline for what correct looks like:
> 
> English (truth)	Spanish (derived)
> text addresses just after a terminator	2,677/2,686 (99.66%)	2,675/2,686 (99.59%)
> consecutive labels one text block apart	2,518 exact	2,660 exact
> pairs the block chain can't reconcile	10	10
> distinct symbols sharing a destination	0	0
> **The honest caveat.** Over half the addresses come from reading a relocated `text_far`, and only **44 of those 2,694 sites sit inside a byte-identical run** — the rest are interpolated. A wrapper is five bytes, so an interpolated site can land next door and read a valid pointer to the _neighbouring label's_ text: a real address, in range, pointing at real Spanish. Nine labels around `_CantDepositLastMonText` did exactly that, each shifted one block, and no range, order or plausibility check noticed. I found it by reading the decoded Spanish against the English and fixed it (the block walk bridges bad anchors, and its exact-closure requirement is what makes the correction safe). So: every check passes and every spot-check is correct, but I'd call it very good rather than proven. A `pokeyellow-es` would still be the thing that proves it.
> 
> Happy to submit this as a follow-up PR on top of #622 — manifest plus the recovery tooling under `tools/es/`, ~1.1 MB for the manifest, same shape as your `_es` ones. Or I can hold it until this lands, or hand it to you to fold in. Whatever you and @bryanthaboi prefer.
> 
> One thing I won't include: the extracted Spanish script itself. That's Nintendo's text and belongs in nobody's repo — the manifest carries addresses and the character map only, exactly like the existing ones.
> 
> Worth adding that the method isn't Spanish-specific. French and German Yellow disassemblies exist but aren't needed for this route; anyone with a European cartridge can generate its manifest the same way.

This is excellent work, and the honesty about the interpolated sites is what makes it credible. My preference: follow-up PR once #622 lands, rather than folding in, the Red/Blue manifests trace to a verified disassembly while yours is cartridge-derived, and each deserves its own review. 

Two asks for the follow-up: include the recovery/verification tooling under tools/es/ so the manifest is reproducible from a dump, and given the _CantDepositLastMonText incident, an on-device play pass through a few interpolated-anchor areas before we flip the UNSUPPORTED entry. You're right about dexUnits, the flag is per-manifest, the generator never decides per entry. @bryanthaboi sound good to you?

hernan0078:
Agreed on all three, and I think you're right about keeping them separate — the provenance really is different. Yours traces to a disassembly that rebuilds the ROM byte-for-byte; mine is inferred from the cartridge and carries a residual risk yours doesn't. Reviewing them together would blur that distinction, and it's the distinction a reviewer most needs to see.

`tools/es/` is already how the manifest is built, so that ask is covered: bank alignment, the recovery strategies with their validation, the character map derived from the font, and the RAM-splice rebuild — plus a README that explains the method and what it costs. One command from two dumps:

```
python3 tools/es/make_es_manifest.py \
    --base "Pokemon - Yellow Version.gbc" \
    --target "Pokemon - Edicion Amarilla.gb" \
    --manifest tools/rom_manifest_yellow.json \
    --out tools/rom_manifest_yellow_es.json --dex-labels AL,PE
```

On verification: I wanted something better than a vibe check, so I tried to derive a risk set first. Two approaches failed and I'll say why, because the failures are informative.

Geometry doesn't discriminate. **2,258 of 2,490** `farptr` labels sit within 12 bytes of a neighbouring wrapper — that's 84% of the dialogue, so "could be off by one wrapper" describes almost everything and sampling it proves nothing.

A corroboration audit — asking which labels a closed block chain independently confirms — returned zero. That was a bug in the audit, not a result: after the full pipeline nearly every label is placed, so there are no gaps left for a chain to bridge. Worth knowing if you ever try the same trick.

What did work is a cross-language anchor: **digits survive translation**, so a line shifted onto its neighbour would carry the wrong numbers. Exhaustively over all 2,694 labels — 194 contain digits in the English, **163 match exactly**, 31 differ. Every one of the 31 I've inspected is correct localisation rather than a shift:

- `5F: DRUG STORE` → `P4: FARMACIA` — Spanish counts `PB, P1…P4`, so every Celadon Mart floor sign is off by one *by design*
- `over 6'8" tall` → `sobrepasa los 2 metros`, `over 44 pounds` → `pesa más de 20 kilogramos`
- `burrow over 60 MPH` → `cava a 90 KPH`

Those last two probably apply to your Red/Blue manifests as well, if you ever diff numerals as a check.

No shifts found by that test. Its limit is that only 7% of labels carry digits — so it's evidence, not proof, and your on-device ask is the right instinct.

Where that stands today: the build is installed on a phone and has been played, not audited. The title, Oak's opening, the Pokédex (entries plus `AL`/`PE` in metric), move and item names, and the PC's messages all came out in correct Spanish, and nothing turned up reading as the wrong character's line. That's genuine on-device evidence, but it was ordinary play rather than a hunt for shifted runs, so I'd present it as such.

If someone does want to go looking deliberately, the three places worth the time are: Celadon Mart's five floors (the tightest consecutive-label group in the flag list, so a shift there shows up as a sign naming the wrong department), a full PC deposit/withdraw/release cycle (where the `_CantDepositLastMonText` run went wrong), and Silph Co 4F/5F (the 26 labels placed by the block walk rather than a pointer read). That's a short route with real diagnostic power, and it needs a Spanish cartridge rather than anything from me — so I'd rather leave when the `UNSUPPORTED` entry comes out to you and @bryanthaboi, and not gate it on a play-through I can't promise a schedule for.

So that it's inspectable rather than described, I've pushed it as a branch on top of this PR's head: **[hernan0078/gen1recomp@spanish-yellow](https://github.com/hernan0078/gen1recomp/tree/spanish-yellow)** — one commit, based on `pull/622/head`, no PR opened. Amarillo becomes the `es` locale of Yellow, the `UNSUPPORTED` entry goes, `tools/es/` lands, and the three packaging lists that name manifests explicitly pick up the new one (iOS derives its list from GameVersion, as you noted).

Checked against your tree rather than mine, since that's the part that matters:

- `tools/build_rom_data.py` completes every stage — 18 data files, 510 assets
- `PIKACHU` comes out `heightM = 0.4`, `weightKg = 6.0`, `heightLabel = "AL"`, `weightLabel = "PE"`, `kind = "RATÓN"` — your Pokédex work needed no change
- moves read `ABSORBER`, `ÁCIDO`, `ARMAD. ÁCIDA`
- `scripts/test.sh --quick`: all tiers pass

Rebase it, cherry-pick it, or ignore it and I'll open a proper PR once this lands — whatever's least work for you.

And yes — the `dexUnits` flag is decided once for the table and written into the manifest; the extractor is told, never left to guess.

bryanthaboi:
While I appreciate the efforts, I think it's more wise to instead consider using this information to help with translation mods. Keeping the rom extraction pinned to the US rom dump I believe is much more simple.

jherediagu:
Fair enough! also good news, I was thinking about as an alternative -> the translation-mod route needs zero engine changes. The mod system already has a first-class text registry (mod.content.text:override, src/mods/Schemas.lua:599), and battle/map dialogue all reads through the merged table. So the shape would be: a small generator tool in the repo that takes the user's own EUR dump plus the existing _es manifests (addresses + charmap only, as before) and emits a personal translation mod locally, extraction stays pinned to the US dumps, no Nintendo text ever enters the repo, and the same shape covers @hernan0078's Yellow work and any other EUR language. If that sounds right to you I'll put it up as a PR: tools/ generator + manifests, nothing touching the import pipeline. @bryanthaboi what do you think?

PR #674, opened 2026-08-02
fix: Oak's intro white field stops short of the dialogue box

spiritsnails:
Oak's speech fills white over the 160x144 UI canvas, but its dialogue box docks to the **window** bottom edge (`Renderer:setUIAnchor`, via `TextBox`). In any letterboxed window that leaves black between the bottom of the white and the top of the box he is speaking from.

Opts into `letterboxWhite`, the same surround a battle already claims, so the voids take the paper shade and the two meet.

Also: the shrink beat redraws the last page itself as a replica at `TextBox`'s own rect while the pic shrinks. That copy now rides the same bottom anchor as the real box it stands in for — otherwise the text jumps up a letterbox on the frame of the swap.

Driver at `tests/drivers/oak_speech_letterbox_test.lua` (resizes to 1000x700 first — an exact multiple of 160x144 has no voids to get wrong).

### DIFF
--- a/src/ui/OakSpeech.lua
+++ b/src/ui/OakSpeech.lua
@@ -20,6 +20,16 @@ local OakSpeech = {}
 OakSpeech.__index = OakSpeech
 OakSpeech.isOpaque = true
 
+-- The speech is a white field with a pic on it, and its dialogue box docks to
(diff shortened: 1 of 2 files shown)


spiritsnails:
Folded into #678, which makes the dynamic UI an option and includes this fix.

PR #675, opened 2026-08-02
Add Celadon Battle Facility: a tiered gauntlet ladder in Kanto

amoore2829:
A new building on Celadon City's west lawn running a three-tier gauntlet: consecutive battles with no healing between rounds, opponents scaled to the player's party, a token currency, a prize counter and a records board.

BRONZE (3 rounds) / SILVER (5, +3 levels) / GOLD (7, +6), each unlocking only once the tier before it is cleared.

Placement is data-verified rather than eyeballed. Celadon is the only oversized city map (25x18 blocks) and block rows 7-11 / columns 1-9 are a contiguous lawn, the largest unused space in Kanto. The footprint is the one 4x2 rectangle there that touches neither a map object (an NPC stands at cell (8,17)) nor the Dept. Store's approach row, which is entered from below and would be walled off by anything on block row 7 columns 4-5.

Two things only the rendered result caught:

- Signage is baked into the tileset art, not drawn as objects, so copying a neighbouring building imports its sign. The first draft put a second "GYM" on the lawn; the second said "MART". Blocks 17 and 115 are the culprits and the suite now asserts neither lands in the footprint.
- The lobby was sized against the lawn (10x9 blocks = 20x18 cells) when the Game Boy screen is 10x9 cells, leaving the whole cast off-camera. It is 7x6 now, with the NPCs on the first row the camera does not clip.

Scaling runs on the engine's trainer.party hook rather than a custom script command: start_battle resolves a static party, so the rosters ship placeholder levels that never reach a battle and the hook rebuilds them per fight. This holds however the battle is reached.

Also gitignore *.gb / *.gbc / *.sav. The ignore rules covered the generated caches but not the ROM itself, so a `git add -A` in a tree with a .gb beside main.lua would have committed a copyrighted ROM to a public repo.

Verified: 340 unit checks, three in-engine drivers (door round-trip, ladder gating and scaling, lobby/records/purchase), modkit validate and lint clean. The engine suite is unchanged.


bryanthaboi:
Mods do not live in this repo

PR #676, opened 2026-08-02
feat: POKEPORT_NO_MODS boots with every mod disabled

spiritsnails:
"Does it still happen with mods off?" is the first question on any bug report from a modded build, and there is currently no way to answer it short of uninstalling or hand-editing options. Joins `POKEPORT_NO_DISCORD` and the rest of the `POKEPORT_*` family. Unset, nothing changes.

```sh
POKEPORT_NO_MODS=1 ./scripts/run.sh
```

A load-time skip, not an uninstall: the manager still lists everything, so you can boot mod-free to confirm, then boot normally and bisect by toggling.

`_saveState` is gated on the flag. It writes an entry for *every* mod, so one toggle in the manager during such a session would have persisted "all off" over the real per-mod choices, destroying the configuration being bisected against.

Docs under Route A in `CONTRIBUTING-mods.md`, since mod authors are the ones who most need "check whether the engine does this too" before filing.

### DIFF
--- a/CONTRIBUTING-mods.md
+++ b/CONTRIBUTING-mods.md
@@ -166,6 +166,30 @@ lowercases and de-dupes. A recommended starting set: `beginner`,
 `data-only`, `quality-of-life`, `hardcore`, `cosmetic`, `story`, `ruleset`,
 `audio`, `ui`, `total-conversion`.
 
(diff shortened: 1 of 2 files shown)


spiritsnails:
Closing for now, focusing on bug fixes.

PR #699, opened 2026-08-02
Add 3X game speed and R2/L2 shoulder button speed hotkeys

### LINKED ISSUE #677: Gamespeed 3x
### One line summary

Title says it all. Please add an option for game speed 3x between 2x and 4x. Would be the perfect bridge option.

### Which game is this about

Not version-specific

_No response_

### What do you want

3x game speed :) (and potentially a button to toggle between selected fast forward speed and default game speed).

### Why is this worth doing

2x is a bit too slow for my taste, while 4x a bit too fast.

### How should it work

Just by adding an option in the main menu. It would also be cool to have a hotkey to switch between fast forward and default speed. :)

### Anything else

_No response_

ShaneMcGovernIE:
### Changes

**3X speed level** — Added `3` between `2` and `4` in `GameSpeed.LEVELS` (`src/core/GameSpeed.lua`). The Options Menu GAME SPEED row now cycles through: 1, 2, 3, 4, 10, 20, 30, 50, 75, 100, 200.

**Controller hotkeys (R2/L2)** — `rightshoulder` cycles speed up, `leftshoulder` cycles speed down. These are handled in `Game:gamepadpressed` before the Input bindings mapping, so they work independently of the player's button configuration without shadowing any GB button.

**Keyboard hotkey (1)** — Pressing `1` cycles speed up, matching the same pattern as hotkey `2` (colors), `3` (tilt), etc.

**Safety gate** — The `_cycleSpeed` helper blocks speed changes during transitions, scripted cutscenes, and link play (identical guard to the color-cycle hotkey at key 2).

### Files changed
| File | Change |
|------|--------|
| `src/core/GameSpeed.lua` | Added 3 to LEVELS list |
| `src/core/Game.lua` | Added `_cycleSpeed` method, keyboard `1` hotkey, gamepad shoulder button handling |

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ supported out of the box.
 | Key       | What it does                                         |
 | --------- | ---------------------------------------------------- |
 | `-` / `=` | Zoom out / in (overworld; also mouse wheel)          |
+| `1`       | Cycle GAME SPEED up (controller: R2 faster, L2 slower) |
 | `2`       | Cycle COLORS                                         |
 | `3`       | Cycle TILT (free-roam overworld)                     |
 | `4`       | Cycle ZOOM through every level (free-roam overworld) |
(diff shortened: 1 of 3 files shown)


bryanthaboi:
Will merge but add hotkey to readme

PR #730, opened 2026-08-03
Hide the touch overlay when a controller is connected

hernan0078:
`noteGamepad` already hides the pad on the first button or stick **use**. But until that press the overlay sits over the screen next to a controller the player has obviously chosen — and on a fresh boot with a pad already paired, it sits there for the whole session until something is pressed.

Connecting a controller is itself the statement of intent, so this hides on connect.

`love.joystickadded` turned out not to be wired anywhere: `main.lua` never defined it, `Boot` did not forward it, and `Game` had no handler. All three are added, plus `TouchControls:joystickadded`.

`joystickremoved` already gave the overlay back. It now respects the same preference, so a player who turned auto-hide off is not overridden by a disconnect — while a controller going flat still returns the pad, since there is nothing else to play with.

**`AUTO HIDE PAD`** in OPTIONS turns it off. Defaults on, and tested with `== false` so an existing `options.lua` without the key keeps the new behaviour rather than needing a migration.

48 lines across 5 files. Never resurrects an overlay the player disabled outright, and never fires where there is no overlay to hide.

Tested on a physical iPhone with a Bluetooth controller: pairing hides the pad mid-game, unplugging brings it back. `scripts/test.sh --quick` passes 73/73 engine and 3/3 modkit on this branch.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -413,6 +413,14 @@ function love.visible(v)
   Game:visible(v)
 end
 
+-- A controller appearing is a statement of intent: the on-screen pad steps
+-- aside for it (TouchControls:joystickadded), and comes back when the last
+-- one is unplugged.
+function love.joystickadded(joystick)
+  if editorMode or TouchEditor or Importer then return end
+  if Game and Game.joystickadded then Game:joystickadded(joystick) end
+end
+
 function love.touchpressed(id, x, y, dx, dy, pressure)
   if editorMode then return end
   if TouchEditor then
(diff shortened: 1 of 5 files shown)


bryanthaboi:
if im gonna roll this in it needs an option in options for when mobile device, MOBILE CONTROLS,  hide on input, hide on controller, never hide, always hide
and perhaps like, shorter game-friendly way to say all of the above in options

PR #759, opened 2026-08-03
Extend ROM-text messages to statuses, items and the learn/evolve flows

jherediagu:
Continues the ROM-text series (da0fa5c, 8c1fbfb): status conditions, trainer-AI item use, item results, evolution and the move-learn flow now print the line the ROM itself carries, with the old engine wording as fallback — so a localized import (e.g. Spanish EUR) shows the ROM's own translation instead of English.

Notable: `_AIBattleUseItemText` restores the real wording ("used X on ONIX!") and skips `prefixEnemy` since the original has no "Enemy " prefix; `_TryingToLearnText` is the whole exchange in pokered, so two engine strings merged into one fallback.

Skipped on purpose (commented in code): `_EvolvedText` (extracts truncated), `_VitaminStatRoseText` (slot order differs per localization), X Accuracy / Guard Spec / Repel (no usable label). A few English sources changed shape, so `tools/modkit.py translation --refresh` will flag orphaned catalog keys.

### DIFF
--- a/src/battle/BattleState.lua
+++ b/src/battle/BattleState.lua
@@ -385,8 +385,8 @@ local function displayName(b)
 end
 
 -- Apply the "Enemy " prefix to a pre-built message from a module that
--- only knows the raw nickname (Status.beforeMove/residual,
--- TrainerAI.useItem): splice it in before the first name occurrence.
+-- only knows the raw nickname (Status.beforeMove/residual): splice it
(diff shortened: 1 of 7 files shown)


bryanthaboi:
please resolve conflicts @jherediagu

jherediagu:
> please resolve conflicts @jherediagu

true! solved! :)

PR #766, opened 2026-08-03
iOS: raise the deployment target so Xcode 26+ can build at all

hernan0078:
On Xcode 26 and later, `build_ios.sh --device` currently fails before compiling anything. Two lines in `patch_love_src.py`.

**The app target** asks for iOS 14.0. Xcode 27 refuses below 15.0 outright.

**liblove** is left at its historical 8.0 deliberately — the comment says `liblove stays as upstream` — and Xcode refuses that too. Because only the app project is patched, the error is reported against `liblove.xcodeproj`, which this script never touches, so it reads as unrelated.

Raised both to 16.0, the value I tested at. 15.0 clears the current floor; 16.0 buys some room before it moves again. Happy to use 15.0 if you would rather stay close to the minimum.

Nothing else is needed on current `dev` — `gr_saveDirectory` already holds its `std::string` in a static, so the only errors left on a clean checkout are the bundle identifier and the HealthKit entitlement, both of which are your project settings rather than bugs.

### Testing

`scripts/test.sh --quick` — 73/73 engine, 3/3 modkit.

`scripts/build_ios.sh --device --release` on Xcode 27 reaches **BUILD SUCCEEDED**, with `GEN1_BUNDLE_ID` pointed at a team I can sign for and the HealthKit entitlement removed locally (neither is part of this PR).

Worth flagging: until this lands, nobody outside the project can build for iOS at all — which is also why [#622](https://github.com/bryanthaboi/gen1recomp/pull/622) has not been tested there.

### DIFF
--- a/mobile/ios/patch_love_src.py
+++ b/mobile/ios/patch_love_src.py
@@ -13,7 +13,7 @@ What it does:
      calls a GR*Bridge Swift class through the Objective-C runtime, so
      liblove never links against Swift directly).
   3. Patches love.xcodeproj so the love-ios app target compiles the native
-     files (Swift 5, iOS 14 deployment for UTType/forExporting APIs).
+     files (Swift 5, iOS 16 deployment -- Xcode 26+ refuses lower).
 """
 
 import re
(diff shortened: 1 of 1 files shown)


castdrian (approved):

castdrian:
ideally please always rebase feature branches just to make sure, there aren't any conflicts here but it's generally good practice to do so 👍🏻

Issue #729, opened 2026-08-03
Support the Spanish (and other localised) Gen 1 releases


hernan0078:
Opening this as a request rather than a PR, because the work is bigger than a patch and the shape of it is yours to decide.

## What players hit today

A Spanish cartridge dump is rejected at import — correctly, since
`GameVersion.forSha1` only knows the three English hashes. So a Spanish
speaker can play the engine, but only in English, and the natural assumption
("I will just import my Spanish ROM") does not work.

I have built an iOS distribution of the engine and this is the single most
common thing people ask me for, ahead of any feature.

## Why this looks tractable

The architecture already treats a ROM release as data rather than as
assumptions in code. `tools/rom_manifest_*.json` carries per-release
`symbols` (4,460 name → [bank, address] pairs), `charmap`, `text`,
`constants` and the rest, and `RomExtractor` reads addresses through
`self:symbol(name)` instead of hardcoding them. Three releases already ship
this way.

So a fourth is, in principle, a fourth manifest — not a fork of the
extractor.

## What would actually be needed

Listing what I found while looking, not prescribing:

1. **A symbol map for the Spanish build.** The current manifests come from
   the English disassembly. Whether pokered/pokeyellow can assemble the
   Spanish releases, or whether the symbols have to come from somewhere
   else, is the part I genuinely do not know and the part everything else
   depends on.
2. **A charmap with the Spanish characters.** The current one is the English
   table; the only accented glyph anywhere is the small `é` of POKéMON.
3. **Font glyphs.** `src/render/Font.lua` has no `Ñ` and no accented vowels,
   so even correct text would render as blanks today.
4. **Label-keyed data.** `dexEntryLabels`, `typeNameLabels` and friends key
   on English label names; those are internal identifiers rather than
   display text, so they may well be fine, but they are worth a look.
5. **`GameVersion` entries** with the Spanish SHA-1s, which is the easy part.

## What I can offer

I cannot commit to the symbol map — that is the real work and I would rather
not promise it than promise it badly. I am happy to do the parts I can
verify: font glyphs for the Spanish characters, charmap work, testing
against real hardware, and the iOS side of it.

If localisation is out of scope for the project, that is a completely fair
answer and worth saying in the README — right now nothing tells a Spanish
speaker that their ROM will be refused before they try it.

bryanthaboi:
[...]

PR #767, opened 2026-08-03
Launcher: route its remaining text through Strings so translation mods can reach it

hernan0078:
You mentioned on #729 that translation mods are the way to go. This removes something in their way: **a translation mod cannot translate the launcher at all today.**

Most of what the launcher draws never reaches the catalog — `GOOD TO GO`, `ROM REQUIRED`, `Verified.`, `SAVE FILES`, `SAVE SLOT`, `LOADED`, `Edit`, `Delete`, `empty slot`, the ROM and save buttons, and the mod browser's chrome are raw literals. So someone who installs a Spanish mod gets Spanish battle text sitting under an English launcher, with nothing they can do about it.

24 sites now go through `Strings()`.

**No behaviour change without a catalog.** `Strings.lookup` returns the source unchanged when nothing is loaded, so a vanilla install draws byte-identical text. This only adds keys a mod may choose to override.

Two of them also had a latent bug worth mentioning: the Edit and Versions buttons measured their width against the English literal while drawing the label, so any longer translation would have overflowed the button. They measure what they draw now.

I found this building a Spanish iOS distribution — I have a working catalog for these keys and I am deliberately not proposing it here, since a built-in table is exactly the thing you said you would rather do with mods. This is just the hooks.

`scripts/test.sh --quick`: 73/73 engine, 3/3 modkit.

### DIFF
--- a/src/import/RomImporter.lua
+++ b/src/import/RomImporter.lua
@@ -3340,9 +3340,9 @@ function RomImporter:_drawGamePanel(version, x, y, w, h, paged)
   printB(gameName, x, y)
   local nameW = self.gameNameFont:getWidth(gameName)
   local pill
-  if ready then pill = { text = "GOOD TO GO", c = PAL.green }
+  if ready then pill = { text = Strings("GOOD TO GO"), c = PAL.green }
   elseif locked then pill = { text = "COMING SOON", c = PAL.disabledInk }
-  else pill = { text = "ROM REQUIRED", c = PAL.gold } end
+  else pill = { text = Strings("ROM REQUIRED"), c = PAL.gold } end
   love.graphics.setFont(self.pillFont)
   local pw = self.pillFont:getWidth(pill.text) + 24 * s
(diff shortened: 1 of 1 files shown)


bryanthaboi:
Conflicts

bryanthaboi:
merged as #791 - thanks for this, the reasoning was right and it's in.

it went conflicting because the launcher panels moved out of RomImporter into LauncherView after you opened it, and most of what you wrapped got routed on the way. of your 22 keys, 11 were already going through Strings on dev and 5 more only matched in comments, so #791 is the six that were genuinely left: ROM, SAVE FILES, SAVE SLOT, Import mod .zip, and the two android hints.

the width bug you found is gone too - chip buttons measure the text they draw now. one related thing did survive: the delete chip pins its width to the unarmed caption, which only works because english Delete is longer than Sure?. japanese さくじょ is shorter than よろしい？, so #791 pins it to the wider of the two.

PR #800, opened 2026-08-04
Pre-add the official mod index to FIND MODS for every install

ShaneMcGovernIE:
## What

FIND MODS currently ships with no default index: the tab opens on an "Add an index" prompt and stays empty until the player names one. This PR pre-adds the project's own index (`bryanthaboi/gen1recomp-mod-index`) for every install, so the tab opens on the official listing right away.

## How it works

- `ModIndex.OFFICIAL` resolves the official index through the same `resolveSource` path players' pastes go through, so feed/base/fallback cannot drift from what `addSource("bryanthaboi/gen1recomp-mod-index")` would store.
- `ModIndex.sources()` tops the returned list with the official index unless the player removed it or already added it by hand (no duplicates, no re-pinning).
- `ModIndex.removeSource` on the official feed records a permanent opt-out (`options.modIndexesOfficialRemoved`), so a player who removes it does not get it back on the next launch. Removing any other index behaves exactly as before.
- Docs (`docs/new-features.md`) and the `RomImporter` FIND MODS comment updated to match.

## Testing

- New `mod_index_tests` section (SaveData stubbed): fresh install lists exactly the official index without persisting anything; player indexes ride below it; a hand-added official index is not duplicated; removal records the opt-out and empties the list; removing a player index leaves the default alone.
- End-to-end smoke: a fresh importer draws the Find panel, picks up the pinned source, fetches the official feed, and lists its mods.
- Launcher regression tests (`launcher_*`, `mod_index_bug597`) pass.

Note: the trust caveat still applies to third-party indexes; the official listing is the project's own and carries the same "listed, not reviewed" notice.

### DIFF
--- a/docs/new-features.md
+++ b/docs/new-features.md
@@ -453,10 +453,14 @@ printout.
 
 A FIND MODS tab sits beside MODS in the launcher and browses a published
 mod index: a metadata-only feed listing mods that live in their authors'
-own repositories. No index ships with the launcher and none is ever added
-automatically, so the tab opens on an "Add an index" prompt until you name
-one; paste an index URL or its `owner/repo` and it is remembered in
-`options.lua`. More than one index can be added, and the listings merge.
+own repositories. The project's own index
+(`bryanthaboi/gen1recomp-mod-index`) ships pre-added for every install,
+so the tab opens on the official listing right away; paste another index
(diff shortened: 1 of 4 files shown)


bryanthaboi:
cant do that but i can make a short url for it

PR #802, opened 2026-08-04
Add Xbox Dev Mode UWP support

caorthann-celt:
## Summary

This adds an Xbox Dev Mode UWP port using LÖVE 11.5, LuaJIT, SDL2 and ANGLE.

The UWP port is contained under `ports/uwp`, with pinned prebuilt dependencies so normal CI builds do not need to fetch or rebuild external projects. Rebuild and verification scripts are included under `scripts/xbox-uwp`.

It adds:

- Xbox UWP CMake and MSIX packaging
- ANGLE rendering
- Xbox controller mappings
- Native Xbox file-picker support for ROMs, saves and mod ZIPs
- LocalState compatible cache handling
- UWP CI and release jobs
- Versioned `.love` payloads shared with the existing release workflow
- Dependency hashes, rebuild scripts and third party licences

## Shared code changes

There are a few small changes outside `ports/uwp`:

- `Platform.lua` now recognises UWP as a console platform.
- The existing iOS native picker state was renamed from `iosPending*` to `pickerPending*` so iOS and UWP can share the same path. The iOS behaviour itself is unchanged.
- UWP avoids the FFI directory helpers in `CacheFs`, which are not suitable for the UWP sandbox.
- The in app update worker is disabled on platforms where networking has not been validated, currently Switch and UWP.
- `scripts/pack_love.sh` can stamp a supplied release version into `Version.lua`.
- Release version calculation was moved into a small shared job so the same version and `.love` payload can be used by the existing builds and the UWP build.

## Signing

PR CI builds an unsigned package and does not require signing credentials.

For the upstream release workflow to produce an installable package, the following **GitHub Actions repository secrets** must be added under **Settings → Secrets and variables → Actions**:

- `XBOX_UWP_SIGNING_CERTIFICATE` — the signing `.pfx` encoded as Base64
- `XBOX_UWP_SIGNING_PASSWORD` — the password for that PFX

The workflow uses these secrets only while signing the package. It automatically derives the public `.cer` from the PFX and includes the `.cer` in the release ZIP. The private PFX is never committed or included in any artifact, and the manifest publisher is taken from its certificate subject automatically.

PowerShell can be used to prepare the certificate secret:

```powershell
[Convert]::ToBase64String(
    [IO.File]::ReadAllBytes("Gen1RecompUWP.pfx")
)


bryanthaboi:
nice u done making updates?

caorthann-celt:
Nice, yeah, I’m done making updates now and everything is passing.  
The only setup needed on your side is for signed releases. Under the upstream repo’s Settings → Secrets and variables → Actions, add:

XBOX_UWP_SIGNING_CERTIFICATE — the signing PFX encoded as Base64  
XBOX_UWP_SIGNING_PASSWORD — the PFX password

PR builds work without them, but the upstream release workflow needs them to produce an installable Xbox package. The public CER is generated automatically, so it doesn’t need a separate secret

bryanthaboi:
already done!

caorthann-celt:
Legend!

PR #829, opened 2026-08-04
Validate save imports by main-data checksum, not raw file size

KikiManjaro:
# Save import: validate by checksum instead of raw size

## Context

While testing the launcher with a Pokémon Red save (a 32812-byte `.sav` exported from VisualBoyAdvance) available at https://projectpokemon.org/home/files/file/5585-pokemon-red-completed-pokedex-with-mew/, I hit a behavior that felt broken: the **SAVE FILES** card systematically refused the file with *"A save file must be 32768 bytes (32 KB); this one is 32812."*.

The problem is that the refusal was purely size-based. A base Gen1 save is exactly 32768 bytes (0x8000), but some emulators append an **RTC footer** of 44 or 48 bytes to the battery image, behind the 32 KB of game data. Result: a usable save gets flagged as invalid just because it is a few bytes too large. I checked the format here: https://bgb.bircd.org/rtcsave.html.

Important detail: on my test save, the first 32768 bytes did carry a valid main checksum (0x98). In other words, it was a real game save, just wrapped in an emulator footer.

### The rule is no longer size, it's the checksum

`SaveFileIO.importToSlot` (src/import/SaveFileIO.lua) no longer just compares `#bytes` against 32768 to decide. It now classifies a non-standard-sized file by the integrity of its **main data**, via the new `GenSave.mainChecksumValid` function (src/save_convert/GenSave.lua).

A Gen1 save's main checksum is the two's complement of the byte sum over the `wMainData` region, stored in a single byte. In our model, that region runs from `0x2598` (start of the player name, `sPlayerName`) to `0x3522` (last byte of `wMainData`, which happens to be the checksum byte itself). I reused the constants already present in GenSave (`O.checksumStart`, `O.checksumEnd`, `O.mainChecksum`) since the export decoder was already using them to *write* the checksum.

`mainChecksumValid` returns one of three states:
- `nil` if the file is too short to even contain the checksum byte (offset 0x3522, so < 0x3523 bytes);
- `false` if the checksum region is present but doesn't match;
- `true` if it matches.

This classification works on any length >= 0x3523 bytes, because the checksummed region fits entirely within the first 0x3524 bytes of a save. So the fate of a truncated or footer-padded file can be decided without a full decode.

### The resulting import policy

- **Oversize (more than 32768 bytes) + valid checksum** → assumed it's an emulator footer. Without `force`, the import doesn't silently happen: it returns `(false, nil, { needsConfirm = true, size })` so the launcher can ask. The player sees an **"Oversized save file"** modal explaining that the extra bytes will be dropped, with an *Import anyway* button. The "yes" branch re-runs the import with `force = true`, truncates to 32768, and the player name, money, and party are preserved. Cancel leaves the file untouched.
- **Oversize + invalid checksum** → rejected as corrupt, never silently truncated.
- **Undersize (fewer than 32768 bytes) + valid checksum** → imported zero-padded. The checksum region is intact, only box banks are missing, which isn't data loss.
- **Undersize + too short to carry the checksum** → still refused with the size message.

On the launcher side: `RomImporter._importSave` passes `force` through to `importToSlot` and builds the confirmation modal. `LauncherView` routes the modal's `kind = "importOversize"` to `_importSave(version, source, true)`.

## Tests

I wanted to verify this at two levels, because the subject (truncating bytes out of the middle of a file) is precisely the kind of thing where a "self-consistent" codec can be wrong on both sides without it showing.

1. **tests/engine/save_file_io_tests.lua** : an always-on policy block, using a real 44-byte MBC3 footer built per the format described on bgb.bircd.org (4 RTC dwords, 16-byte latched copies, 8- and 4-byte timestamps).

2. **tests/save_oversize_vendor_test.lua** : the oversize fixture is built by the **vendor** codec `tools/save_convert/vendor/gen1lib.lua`. The post-truncation export is then re-parsed by that same vendor codec to prove only the footer was dropped (name, money, trainer id, empty party, and 12 boxes all preserved).

## Before 

https://github.com/user-attachments/assets/b6bcb7d6-5134-4884-a810-dcf6c0f76f9b

## After

https://github.com/user-attachments/assets/d3dc0cb8-ade8-44a8-884b-a4dd4212d100

Note: nothing sensitive was committed, i've used many .sav files to be sure that the cycle was correct.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -22,6 +22,7 @@ set -uo pipefail
 cd "$(dirname "$0")/.."
 
 LUA=${LUA:-luajit}
+LUA54=${LUA54:-lua5.4}
 BLESS=0
 QUICK=0
 SHOTS=${WITH_SHOTS:-0}
@@ -134,6 +135,16 @@ if [ -f data/generated/maps.lua ]; then
     run_tier "T3 save editor: wheel scrolling" "$LUA" tests/save_editor_wheel_bug595_test.lua
     run_tier "T3 save editor: pad / NX input" "$LUA" tests/save_editor_pad_input_test.lua
     run_tier "T5 link (loopback lockstep)" "$LUA" tests/run_link_tests.lua
+    # The oversize-save vendor oracle (tests/save_oversize_vendor_test.lua)
+    # cross-checks the launcher's footer-truncation import against the
+    # INDEPENDENT PKHeX-derived gen1lib codec, which cannot run under luajit
+    # (native 5.3+ operators).  Needs a stock Lua 5.3/5.4; skip when absent.
+    if command -v "$LUA54" >/dev/null 2>&1; then
+      run_tier "T3 save oversize vendor oracle" "$LUA54" tests/save_oversize_vendor_test.lua
+    else
+      echo ""
+      echo "-- T3 save oversize vendor oracle: skipped (no '$LUA54' on PATH; set LUA54=...)"
+    fi
   fi
 else
   echo ""
--- a/src/import/LauncherView.lua
+++ b/src/import/LauncherView.lua
@@ -1505,6 +1505,8 @@ local function buildConfirmModal(imp, m)
           imp:_confirmModUpdate(c.id, c.release)
         elseif c.kind == "enableAll" then
           imp:_setAllMods(true, true)
+        elseif c.kind == "importOversize" then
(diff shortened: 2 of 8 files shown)


bryanthaboi:
Conflicts

KikiManjaro:
@bryanthaboi conflicts fixed

image

PR #836, opened 2026-08-05
Add haptic feedback to the on-screen touch controls

### LINKED ISSUE #806: Haptic Feedback for On-screen controls (mobile)
### One line summary

The On-screen controls are neat but a feature that would make them easier to memorize and/or find is haptic feedback.

### Which game is this about

Not version-specific

thegamerninja556

### What do you want

It should be an option that can be easily toggled on/off and maybe allow for haptic feedback intensity. The player would be able to feel the controls be pressed and make it much harder for them to be missed.

### Why is this worth doing

This is worth doing because almost every emulator or way to play retro games on mobile offers this.

### How should it work

- Option in the Settings Menu
- Toggled ON/OFF
- (possibly) Variable intensity of vibrations

### Anything else

This project is very nice and the development efficiency is unlike any other, It lack some small things like this and with them it would arguably be the best way to play Generation 1.

KikiManjaro:
# Add haptic feedback to the on-screen touch controls

## Context

Cf. https://github.com/bryanthaboi/gen1recomp/issues/806

### A single HAPTIC LEVEL option, off by default

Options gains one row, **HAPTIC LEVEL**: an integer **0–10** (0 = OFF, the default). It scales the vibration duration from a barely-there 15 ms tick up to a 400 ms buzz.

### Every fresh press pulses

`TouchControls.pressBtn` now calls `Haptics.pulse()`, it fires on a fresh press only.

### OEMs mutint with Android

LÖVE's stock `GameActivity.vibrate()` used the deprecated `Vibrator.vibrate(ms)`, which the system routes as `USAGE_UNKNOWN` and some OEMs (MIUI/HyperOS) set to **OFF** in their vibration-intensity settings, scaling the amplitude to zero while the call still "succeeds". Choosed to use `VibrationEffect.createOneShot(ms, 255)` with `AudioAttributes.USAGE_GAME` at max amplitude.

## Video

https://github.com/user-attachments/assets/543efd13-ec95-4b49-a3b1-7940647fb209

## Misc

Tested on Android (on a Xiaomi 13 (HyperOS))

Closes https://github.com/bryanthaboi/gen1recomp/issues/806

### DIFF
--- a/mobile/android/love/src/main/java/org/love2d/android/GameActivity.java
+++ b/mobile/android/love/src/main/java/org/love2d/android/GameActivity.java
@@ -53,12 +53,15 @@ import android.hardware.Sensor;
 import android.hardware.SensorEvent;
 import android.hardware.SensorEventListener;
 import android.hardware.SensorManager;
+import android.media.AudioAttributes;
 import android.media.AudioManager;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Environment;
 import android.os.Handler;
 import android.os.Looper;
+import android.os.VibrationEffect;
 import android.os.Vibrator;
 import android.util.Log;
 import android.util.DisplayMetrics;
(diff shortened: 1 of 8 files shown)


bryanthaboi:
Should be gated to controllers, iPhone and Android and added to all use case if so. I don't want to add if only for Android

KikiManjaro:
Thanks for the feedback! I think there may be a small misunderstanding about the scope of this change. The idea here is not to add controller rumble, but specifically to provide haptic feedback for the on-screen TouchControls when pressing virtual buttons.

I totally understand your concern about keeping the feature consistent across platforms. I’m currently looking into how we could provide the same behavior on both Android and iOS, rather than having an Android-only implementation.

PR #871, opened 2026-08-05
Ship a Linux arm64 (aarch64) AppImage

ratherDashing:
## What

Adds `gen1recomp-<version>-linux-arm64.AppImage` to every release, for 64-bit ARM desktop Linux — Raspberry Pi 4/5 on Raspberry Pi OS, Armbian and other SBC distros, arm64 VMs on Apple Silicon.

Today `scripts/build.sh linux` only produces x86_64. It works by unpacking LÖVE's official `love-11.5-x86_64.AppImage`, dropping `game.love` in, flipping on the `FUSE_PATH` hook in `AppRun`, and gluing the squashfs back onto the runtime. There is no aarch64 image to unpack the same way: **LÖVE 11.5 ships win32, win64, macOS, Android, iOS and exactly one x86_64 AppImage**, and that is the whole asset list. So arm64 Linux had no artifact at all.

## How

`scripts/build_linux_arm64.sh` compiles LÖVE 11.5 from the official `linux-src` tarball inside a **Debian bullseye arm64** container and assembles the AppImage from scratch, then fuses it onto a pinned AppImage type-2 runtime. Layout and the `AppRun` contract mirror LÖVE's own x86_64 image, so it is the same idea `build.sh` already uses, one level lower.

Both pinned inputs are SHA-256 verified **on the host** before the container starts, and the container itself runs with no network access.

### Why bullseye, when nobody is asking for bullseye

Bullseye is the *compile environment*, not a claim about where the artifact runs. glibc is backward compatible but not forward compatible, so linking against the oldest supported glibc is the only thing that lets one artifact work everywhere; building on a newer distro silently raises the floor and strands users, with no symptom until they download it.

The binaries come out needing only **glibc 2.29** and **GLIBCXX_3.4.21** — comfortably covering Raspberry Pi OS bullseye → trixie and Ubuntu 20.04 onward. CI fails the build if that floor ever climbs above 2.31.

### The bug worth reading about

The dependency walker bundles LÖVE's own libraries and leaves three categories to the host: driver-coupled (GL/Mesa, X11/Wayland, PulseAudio/ALSA), loader-coupled (glibc, libstdc++), and the font/compression stack (freetype, fontconfig, libpng, brotli, zlib).

That third one is not tidiness. Debian's `libtheoradec.so.1` is — oddly — linked against `libcairo.so.2`. LÖVE needs theora for `love.video`, so the **host's** cairo gets pulled into the process. The dynamic loader resolves one SONAME exactly once per process, so that host cairo then binds to whatever `libfreetype.so.6` *we* bundled:

```
love → liblove → libtheoradec → libcairo (host, new)
                                   ╰→ FT_Get_Transform → libfreetype (ours, bullseye 2.10.4)
```

`FT_Get_Transform` arrived in FreeType 2.11, so cairo 1.18 on a trixie host fails to relocate and the game dies at startup:

```
bin/love: symbol lookup error: /lib/aarch64-linux-gnu/libcairo.so.2: undefined symbol: FT_Get_Transform
```

Bundling a *newer* freetype only moves the arms race along one release. Excluding the whole font stack instead makes the process self-consistent: cairo, fontconfig and freetype all come from one host and agree with each other, while `liblove` — compiled against 2.10.4 — only ever asks for symbols every supported host already has.

## CI

Three path-gated jobs, following the Switch/Xbox UWP pattern already in `ci.yml`:

| job | runner | what it does |
|---|---|---|
| `linux-arm64-selftest` | `ubuntu-latest` | Offline gate: pins are real digests on a **dated** tag (not the moving `continuous`), Dockerfile still on bullseye, exclude list still classifies known sonames correctly, `AppRun` still launches `game.love` with `--fused`, and the host-arch guard actually fires. No container, no arm64 machine. |
| `linux-arm64-build` | `ubuntu-24.04-arm` | Real build, then extracts the artifact and asserts the layout, that every bundled object resolves under `AppRun`'s `LD_LIBRARY_PATH`, and that the glibc floor is still ≤ 2.31. Uploads the AppImage. |
| `linux-arm64` (release) | `ubuntu-24.04-arm` | Reuses the shared `game.love` from `love-payload` so every platform ships identical bytes; staged and published like the other assets. |

Unlike `switch-build`, none of this needs secrets or self-hosted hardware — GitHub's arm64 runners are free for public repos — so **it runs on fork PRs too**.

## Testing

Built and played end to end on a **Raspberry Pi 5, Debian 13 trixie, Wayland**, with a real Yellow cartridge dump:

- **ROM import works from inside the AppImage.** Dropped in a US Yellow dump (SHA-1 `cc7d0326…`); the importer routed it by SHA-1, auto-selected the Yellow tab, decoded it, and wrote the generated cache to `~/.local/share/pokemon-love2d/yellow/` — i.e. the normal per-user location, not somewhere inside the read-only squashfs mount.
- **It plays.** Oak's intro through to the overworld in his lab: tiles, NPC and Pikachu sprites, and the dialogue box all render correctly.
- **Audio works.** PulseAudio shows a live uncorked sink input from `love`, so bundled OpenAL is talking to the host's sound server — which is the bet the exclude list makes (bundle OpenAL, leave `libpulse` to the host).
- Fonts render, which is the other half of the freetype decision: `love.font` resolves against the host's freetype and the host's cairo stays happy.

Build and static checks:

- full build from a clean cache: ~3.5 min on a Pi 5; rebuilds reuse the cached LÖVE prefix and take seconds
- two independent clean-cache builds produced the same 13-library bundle set and the same glibc 2.29 floor (the AppImages are not bit-identical — squashfs and the compile embed timestamps — so this is repeatability, not byte reproducibility)
- bundled library set matches LÖVE's own AppImage manifest, minus the font stack
- `--appimage-extract` works without FUSE, so the CI verification runs on runners with no `/dev/fuse`
- `scripts/linux-arm64/selftest_build_linux_arm64.sh` passes
- `scripts/test.sh` (all ROM-free tiers) and `tests/switch_ci_workflows_test.lua` still pass
- `shellcheck -S warning` clean on all new scripts

## Notes for review

- Nothing existing changes behaviour: `scripts/build.sh` is untouched, and the new script is a sibling of `build_switch.sh` / `build_android.sh` rather than another `build.sh` target, because it requires an aarch64 host and the release runner is a Mac.
- Docs: `docs/linux-arm64-build.md` (player instructions, host requirements, the bullseye rationale, the freetype story, how to bump the pins) + a short README section.
- Pins live in one place, `scripts/linux-arm64/common.sh`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -330,10 +330,24 @@ jobs:
 
           # Every bundled object must resolve once AppRun's LD_LIBRARY_PATH is
           # applied; an unresolved soname here is a user-visible launch crash.
+          #
+          # This runs on a HEADLESS runner on purpose, and that is the point.
+          # The first version of this build bundled Debian's SDL2, which
+          # hard-links libpulse/libasound/libX11/libwayland, so it only ever
+          # started on a full desktop -- a bare runner is what exposed it.
           missing="$(LD_LIBRARY_PATH="$PWD/squashfs-root/lib" \
             ldd squashfs-root/bin/love squashfs-root/lib/*.so* 2>/dev/null \
             | grep 'not found' || true)"
           [ -z "$missing" ] || { echo "::error::unresolved deps:"; echo "$missing"; exit 1; }
+
+          # Nothing may hard-link a driver, session or audio-stack library:
+          # those must be reached through dlopen so the AppImage runs on a box
+          # with only ALSA, only Wayland, or only KMSDRM.
+          linked="$(for f in squashfs-root/bin/love squashfs-root/lib/*.so*; do
+            objdump -p "$f" 2>/dev/null | awk '/NEEDED/{print $2}'
+          done | sort -u | grep -E '^lib(pulse|asound|X11|wayland|GL|EGL|drm|gbm|xcb|cairo|sndio|dbus)' || true)"
+          [ -z "$linked" ] \
+            || { echo "::error::these must be dlopened, not linked:"; echo "$linked"; exit 1; }
 
           # The whole point of compiling on bullseye. If a future change moves
           # the builder to a newer base, the glibc floor silently rises and
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -260,6 +260,103 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+  linux-arm64-changes:
+    name: detect Linux arm64 changes
+    runs-on: ubuntu-latest
+    outputs:
+      changed: ${{ steps.paths.outputs.changed }}
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+      - id: paths
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          HEAD_SHA: ${{ github.sha }}
+        run: |
+          if [ -z "$BASE_SHA" ] || [ "$BASE_SHA" = "0000000000000000000000000000000000000000" ]; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if git diff --name-only "$BASE_SHA" "$HEAD_SHA" | grep -Eq '^(scripts/build_linux_arm64\.sh$|scripts/linux-arm64/|scripts/pack_love\.sh$|docs/linux-arm64-build\.md$|\.github/workflows/(ci|release)\.yml$)'; then
+            echo "changed=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "changed=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  linux-arm64-selftest:
+    name: Linux arm64 offline selftest
+    needs: linux-arm64-changes
+    if: needs.linux-arm64-changes.outputs.changed == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      # Deliberately on x86_64: everything this gate checks (pins, the
+      # host-arch guard, the dependency exclude list, the AppRun fusion
+      # contract) is answerable without a container or an aarch64 machine,
+      # so the slow native job below only ever starts on a sane tree.
+      - name: Linux arm64 offline selftest
+        run: bash scripts/linux-arm64/selftest_build_linux_arm64.sh
+
+  linux-arm64-build:
+    name: Linux arm64 AppImage build
+    needs: [linux-arm64-changes, linux-arm64-selftest]
+    if: |
+      always()
+      && needs.linux-arm64-changes.outputs.changed == 'true'
+      && needs.linux-arm64-selftest.result == 'success'
+    # No fork restriction, unlike switch-build: this needs no secrets and no
+    # self-hosted hardware, just GitHub's free arm64 runner for public repos,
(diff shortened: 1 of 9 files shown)


bryanthaboi:
failing test

ratherDashing:
The `libpulse.so.0 => not found` failure is fixed, and it turned out to be a real portability bug rather than an over-strict check. Pushed as f92364a.

**What the headless runner caught.** Debian's `libSDL2` lists `libpulse`, `libasound`, `libX11` and `libwayland-client` as `DT_NEEDED` — resolved by the loader at startup, not dlopened — so an AppImage bundling it refuses to start unless the host has all four. It passed local testing only because a desktop Pi has all four; a bare runner is what exposed it. Debian's OpenAL does the same via `libsndio`, which itself hard-links `libasound`.

Fixing that surfaced a second one. With OpenAL correctly dlopening ALSA, ALSA's config loads its PulseAudio hook plugin, which pulls the **host's** `libsndfile` into the process. `libsndfile` links `libogg`, `libvorbis` and `libmpg123` — the same three that were bundled — and since the loader resolves a SONAME once per process, it bound to the bullseye copies:

```
openal -> libasound -> libasound_module_conf_pulse -> libsndfile (host, new)
                                                         `-> mpg123_info2 -> libmpg123 (ours, bullseye 1.26)
```

`mpg123_info2` arrived in mpg123 1.32, so the plugin failed to relocate, ALSA config collapsed, and the game ran with **no audio device at all** — silently, no crash.

**The fix.** SDL2, OpenAL, libtheora, libogg, libvorbis and libmpg123 are now built from source alongside LÖVE, configured to dlopen their backends. This reduces to one rule, now enforced by an assertion in `build_appimage.sh` rather than a comment: *never bundle a library the host's own stack may also load unless ours is at least as new as theirs.*

Host requirements drop from "a working desktop" to:

```
glibc 2.29+   libstdc++   libfreetype6   zlib
```

Everything else — OpenGL/Mesa, X11, Wayland, KMSDRM, ALSA, PulseAudio — is dlopened, so one image now covers a full desktop, a Wayland-only session, and a KMSDRM handheld with no X server. Bundled libraries went 13 -> 10; `libcairo`, `libpixman` and `libsndio` are gone entirely. Cold builds take ~10 min (six libraries plus the engine); cached rebuilds are seconds.

**On CI status.** This PR's own run is sitting at [`action_required`](https://github.com/bryanthaboi/gen1recomp/actions/runs/31039303518) pending maintainer approval, so it has not executed and reports nothing. To avoid guessing, I ran the identical jobs on the same commit in my fork, where no approval gate applies — **all checks green, 0 failing**:

https://github.com/ratherDashing/gen1recomp/actions/runs/31039693886

```
detect Linux arm64 changes      success
Linux arm64 offline selftest    success
Linux arm64 AppImage build      success   (3m13s on ubuntu-24.04-arm)
```

That is a fork run, not this repo's — worth approving the pending run here to confirm independently. The build job's verification step (the step that failed last time) passes: layout, every bundled object resolving under `AppRun`'s `LD_LIBRARY_PATH`, nothing driver/session/audio hard-linked, and the glibc floor still at 2.29.

Also re-verified by hand on a Raspberry Pi 5 (Debian trixie, Wayland) with a real Yellow dump: boots, imports the ROM, plays, and audio works. Incidentally SDL 2.30 now picks the native Wayland backend instead of falling back to XWayland as bullseye's 2.0.14 did.

PR #927, opened 2026-08-06
Mod API: expose Android second-screen touch input

AverageConsumer:
## Summary

- queue touch input from the existing Android secondary `Presentation`
- expose it to `render.compose` mods through `ctx.secondScreen.pollTouch()`
- report coordinates in the submitted frame's pixel space and clear pending input when the display is torn down
- keep older native bridges safe: `pollTouch()` returns `nil` when the symbol is unavailable

## Motivation

`render.compose` already lets a mod draw a companion UI on a second physical display, but that display is currently output-only. This adds the missing input half without changing normal rendering, primary-screen input, or display selection. Kanto Gear needs it for its touch UI, and the bridge is generic for any second-screen mod.

`pollTouch()` returns the oldest queued `action,x,y` event (`down`, `up`, or `cancel`) or `nil`. The bounded queue emits `cancel` on overflow so a mod cannot be left with a stuck press.

## Verification

- `luajit tests/engine/second_screen_touch_test.lua` — 3/3 checks passed
- `assembleEmbedNoRecordDebug` — passed, including Java and NDK C++ compilation
- the full engine runner reaches and passes the new suite; its three unrelated Windows-only failures were reproduced unchanged on clean `upstream/dev`

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -221,10 +221,11 @@ the finished `worldCanvas` and `uiCanvas` with their SGB `zones` / `worldZones`,
 `worldActive`, the frame metrics (`ww`, `wh`, `pw`, `ph`, `ox`, `oy`, `vpw`,
 `vph`, `scale`, `Sx`, `Sy`, `dpiX`, `dpiY`), `renderer:blitCanvas(...)` for a
 palette-correct blit of either canvas into an arbitrary screen rect, and the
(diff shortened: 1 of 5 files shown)


bryanthaboi:
github was broken when u made this can u push a empty commit to retrigger jobs

PR #940, opened 2026-08-07
Allow TitleState player overrides

### LINKED ISSUE #933: Add TitleState override for player
### One line summary

Mod API for replacing the player character seen on the default TitleState.

### Which game is this about

Not version-specific

_No response_

### What do you want

Currently the TitleState has the path "assets/generated/title/player.png" hardcoded directly and it's difficult to modify through mods without doing a Total Conversion title screen replacement. Please provide an easy method for mods to replace this piece of art, similar to how the title logo can be replaced.

### Why is this worth doing

This is a significant piece of replacing the player character in a mod or texture pack.

### How should it work

- Utilize the field.title.player / boot.title.player registry value (currently unused)
- Add "overrides/" support for all "assets/generated/title/" paths generally by adding Assets.resolve() to the tryImage function in TitleState. [done 5bc7d203a50df6e39a2595c224eb9d051a428eb0]
- If the title player becomes animated at a later date (eg. to "toss the ball" when the TitleMon cycles), fall back to a static image for mods which replace it by default, and then consider adding a separate registry field dictating the ball's animation offsets (ex. the "hand location") to allow mods to incorporate the animation properly if desired.

### Anything else

_No response_

Yukitty:
Closes #933

### DIFF
--- a/src/ui/TitleState.lua
+++ b/src/ui/TitleState.lua
@@ -176,7 +176,8 @@ function TitleState.new(game, opts)
   self.versionFull = imagePath(title.versionRibbon) ~= nil
   self.version = tryImage(imagePath(title.versionRibbon or title.version)
                           or "assets/generated/title/red_version.png")
-  self.player = tryImage("assets/generated/title/player.png")
+  self.player = tryImage(imagePath(title.player)
+                         or "assets/generated/title/player.png")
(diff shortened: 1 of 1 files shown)


Yukitty:
Rebased (now that Assets.resolve is used in the title screen)

Issue #852, opened 2026-08-05
Cannot switch between saves states on smaller 4:3 screen or in vertical mode


cmcgpgt-maker:
### One line summary

On the main menu where you pick your game and save states, the menu for the save states is now hidden by the lower three buttons. This happens on a 4:3 square screen. Originally the save state option were off to the right side.

### Which game were you playing

Red

### Which build are you running

Android

### Were any mods on

Yes

### Which mods (if any were on)

dramatic shape 1.6.0

### Version

1.6.9

### Discord username (optional)

cmcgpgt

### Where in the game

main menu of the app on vertical  mode with or turn either way with a more square resolution device

### Screenshot, or a detailed description if you couldn't take one

<img width="720" height="1560" alt="Image" src="https://github.com/user-attachments/assets/76f2958a-fce5-45b9-89c5-4cbe0bbea3c7" />

### How to make it happen

1. open the app 
2. scroll to the bottom 
3. try to pick a stave state
4. set your screen to vertical mode
5. otherwise have a device with a squarish resolution

### What should have happened

The same state option needs to move back to the right or have a scroll option with the touch screen.

### Anything else

I want to thank you so much for this and let you know how thankful I am the work you put into this.

PR #954, opened 2026-08-07
feat(launcher): dynamic fit-scaling for short and square windows

ShaneMcGovernIE:
## Summary

On short or square windows the launcher stopped at the fixed 0.9 scale floor and let content sit under the fold — the RG34XX's 720×720 screen rendered one column with a 353px page scroll and the footer off-screen. This PR adds **fit-scaling**: the whole UI (text, tap targets, padding, row heights) shrinks together off `Kit.scale` until the panel minimum plus the header/footer chrome lands on screen, instead of scrolling.

## How it works

- `Kit.layout(w, h, fitH)` — an optional third argument solves for the scale at which the caller's panel minimum (unscaled px) plus chrome fits the window height. `FIT_FLOOR = 0.78` is the legibility floor; below it the existing page scroll remains the backstop.
- `Layout.metrics(maxAppW, fitH)` — resolves the column count against the resolution scale first, then fit-scales against the matching per-column minimum. Because fit-shrinking can flip the column count (a 720px square drops below the two-column breakpoint), it re-fits against the new minimum until scale and columns stabilise.
- `LauncherView` — the launcher's column-aware panel minimums (#852: 460 two-col / 660 one-col) now double as the fit-scale target and are passed as the fit hint.
- Fonts rebuild when the fit changes scale at a fixed resolution (font cache key now includes `s`).

## Effects

- **720×720 (RG34XX):** two-column, scroll-free, footer and seal on screen (was one-column + 353px scroll).
- **480×900 / 360×780 (portrait phones):** whole one-column stack fits; no scroll (was unreachable slot list).
- **1280×800 desktop:** unchanged — fit only kicks in when the resolution scale alone would put content under the fold.
- **Below the floor (e.g. 480×360):** still scrolls and clamps to a reachable bottom.

## Tests

`tests/engine/launcher_one_column_reach_bug852.lua` rewritten for the new mechanism: 480×900 and 360×780 now assert `scrollMax = 0`; a below-floor 480×360 window still engages the page scroll and the wheel clamps to the extent. `scripts/test.sh --quick`: **ALL TIERS PASSED**.

Docs updated in `docs/launcher.md` (Responsiveness → Fit scaling).

## Files

- `src/ui/kit/Kit.lua` — fit-scale solve, `FIT_FLOOR`/`FIT_CHROME`, font key includes scale
- `src/ui/kit/Layout.lua` — `metrics(maxAppW, fitH)` + `columnCount` extraction + stabilisation loop
- `src/import/LauncherView.lua` — panel minimums as the fit hint
- `tests/engine/launcher_one_column_reach_bug852.lua`
- `docs/launcher.md`

### DIFF
--- a/docs/launcher.md
+++ b/docs/launcher.md
@@ -223,8 +223,9 @@ filesystem stub the slot backend uses (`tests/engine/save_file_io_tests.lua`).
 ## Responsiveness
 
 Every measurement derives from `love.graphics.getDimensions()` each frame
-plus the existing global scale `s = clamp(height / 768, 0.7, 1.6)`; nothing
-assumes a fixed window size. The game panel's two-column grid (ROM/SAVE
+plus the existing global scale `s = clamp(min(width / 640, height / 768),
+0.9, 1.6) * 1.3`; nothing assumes a fixed window size. The game panel's
+two-column grid (ROM/SAVE
 FILES/Play on the left, SAVE SLOT on the right) collapses to one stacked
 column, slot card below Play, when the window is too narrow for both
 `~300 * s`-wide columns. The save-slot list and the mod list both scroll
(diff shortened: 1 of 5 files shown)


bryanthaboi:
there are conflicts to be resolved and id appreciate screenshots of normal sized windows, small screens, and the one u built this for

bryanthaboi:
conflicts

castdrian:
this needs a rebase

Issue #17, opened 2026-07-21
Incorrect Character Visuals (Only with GBC filter)


SegaSaturnFan:
### One line summary

Characters only have 3 colors (including transparency) when they should have 4

### Which build are you running

Windows

### Version

0.1.5

### Where in the game

Everywhere

### Screenshot, or a detailed description if you couldn't take one

<img width="609" height="488" alt="Image" src="https://github.com/user-attachments/assets/f8c7e612-f243-4e48-ad88-456cea75f6c1" />

### How to make it happen

Simply go anywhere where you can see your player/NPCs

### What should have happened

Not including transparency, NPC and the player sprites should have three colors

### Anything else

Appears to only be an issue in the windows version

bryanthaboi:
get latest version and try new color palete called red, let me know if you like that one

SegaSaturnFan:
The new color palette is good, reminds me of crystal version, I like it, as for the GBC filter, it still technically isn't accurate, I think it's just an issue with that one filter not setting the NPC colors correctly, all the other filters seem to set it correctly, i'll edit the issue name to reflect this better

Issue #15, opened 2026-07-21
you shouldn't be able to black out against blue in firrst match


bryanthaboi:
the first match against blue, you should not black out if you lose. the game is supposed to auto heal your party after this match.

Issue #14, opened 2026-07-21
non-main gym trainers that you have not fought should deactivate after defeating gym leader and earning badge


bryanthaboi:
In the original game, if you earn a gym's badge without defeating all of that gym's trainers beforehand, the unfought trainers will deactivate

Issue #12, opened 2026-07-21
Pokemon not getting evolution moves


SegaSaturnFan:
### One line summary

When a pokemon evolves and has an evolution move, it does not get that move, (I.E. Kakuna not getting harden)

### Which build are you running

Windows

### Version

0.1.4

### Where in the game

Anywhere where pokemon can evolve

### Screenshot, or a detailed description if you couldn't take one

<img width="1123" height="980" alt="Image" src="https://github.com/user-attachments/assets/23225920-dc86-4e79-a55a-e0dfbea4a435" />

### How to make it happen

Simply Evolve a pokemon that's supposed to get an evolution move

### What should have happened

upon evolving, the pokemon will gain the evolution move

### Anything else

Unsure if this is a specific issue with the battle i was in before hand or if evolution moves are just broken, further testing is recommended

SegaSaturnFan:
Not an issue, my mistake

sebastian926:
I wanted to reopen this, because it is an issue. In the original game Pokemon DO learn moves at the level their evolutions do when evolving: https://youtu.be/OI2lS6aEC1g?t=1156

I was casually using Magikarp and evolving it to Gyarados in level 20 in this run and he didnt got bite: 

<img width="1172" height="1047" alt="Image" src="https://github.com/user-attachments/assets/6807a57d-d029-4309-919a-6ea08d80302a" />

Different would be if he evolved at level 21, then he wouldnt get bite.

Poor Gyarados, already two attacks skipped, lol.

SegaSaturnFan:
> I wanted to reopen this, because it is an issue. In the original game Pokemon DO learn moves at the level their evolutions do when evolving: https://youtu.be/OI2lS6aEC1g?t=1156
> 
> I was casually using Magikarp and evolving it to Gyarados in level 20 in this run and he didnt got bite:
> 
[...]

PR #965, opened 2026-08-07
Add mod.world:startWildBattle

crusty:
`src/world/WorldAPI.lua` states its own contract in the header:

> `mod.world`: the supported way for mod code to act on the running overworld. … Reaching into OverworldState internals stays unsupported; anything a mod legitimately needs belongs here.

Starting a wild encounter isn't there. A mod that wants one has to build a `BattleState` and push it, and the two things that matter are invisible from the outside: `battle.onFinish` drives `OverworldState:afterBattle`, which is the only caller of `Evolution.checkParty` and of the blackout-on-loss path, and `pushBattle` owns the entry wipe and the battle theme. Neither raises when it's missing. The mon gains EXP, levels normally, and simply never evolves.

## Why now

A published mod hit exactly this. Players lost every level-up evolution for days with no error to go on, and a party wipe left them standing on the map instead of blacking out to the last heal point. The fix has been written three times over and none of it is in the shipped release.

I'm not asking this repo to take responsibility for a third-party mod. The relevant part is that the failure mode is completely silent and the engine currently offers no supported path that would have made it impossible — which is the gap `WorldAPI` exists to close.

## What's here

`mod.world:startWildBattle(species, level)` resolves the live overworld, validates its arguments, wires `onFinish` to `afterBattle`, and routes through `pushBattle`. It follows the file's existing conventions: `nil, "no overworld"` when none is up, `nil, reason` on bad input, `true` on success.

It deliberately does **not** return the `BattleState` — handing a mod that object is the internals-reaching the header rules out.

## Refusal contract

Worth reviewing explicitly, because two of these are judgement calls rather than plain validation:

| Condition | Returns |
|---|---|
| no overworld up | `nil, "no overworld"` |
| unknown species | `nil, "unknown species: X"` |
| level not a whole number 1..100 | `nil, "level must be a whole number 1..100"` |
| no healthy party | `nil, "no healthy party"` |
| a battle or battle transition already on the stack | `nil, "a battle is already running"` |
| world mid-warp | `nil, "the world is mid-warp"` |

**The re-entrancy refusal is new behaviour, not just hardening.** `WorldAPI:overworld()` deliberately resolves the world from *under* whatever sits on top of it, so without this a call from a battle hook stacks a second battle over the live one — and on a loss, the inner `afterBattle` heals the party, halves money and warps to the heal point while the outer `BattleState` is still on the stack. Refusing forecloses a mod deliberately queuing a battle from inside another; I think that trade is right, but it is the reviewer's call.

**The healthy-party test runs before `newWild` is constructed**, not after. `BattleState.newWild` calls `markSeen(game, species)` (`BattleState.lua:596`) before a caller can inspect `battle.dead`, so checking afterwards means a refused `startWildBattle("MEWTWO", 70)` still permanently flags MEWTWO as SEEN in the save.

The validation isn't redundant with what's downstream either: `Pokemon.new` asserts on an unknown species, which is a throw rather than this API's `nil, reason` contract, and it never checks level at all — `tonumber` accepts `5.5`, which is written straight into `mon.level`, `Stats.calc` and the exp curve.

## Test

Folded into `tests/mod_world_tests.lua`, which is already the `mod.world` suite, already T3, and already carries the same off-the-world refusal pattern for `warpTo` / `spawnNpc` / `queueScript`. It reuses that file's `liveWorld` fixture rather than standing up a second one. 17 new checks; the suite goes 19083 -> 19100.

Coverage: the argument refusals above, then the handoff end to end -- the entry transition is pushed first, it hands off to the battle, `awardExp` records `leveledUp`, and the win reaches `EvolutionState`.

That last chain had no coverage. `tests/parity_trainer_evolution_order.lua:32-33` replaces `BattleState` wholesale with `{ leveledUp = { [caterpie] = true } }`, so `awardExp -> leveledUp -> afterBattle -> checkParty` was never exercised against the real thing.

T3 rather than the ROM-free tier is forced: the handoff only exists once a real map is up (`OverworldState:setMap` needs real tilesets and the bound `Game` upvalue), and the fixture dataset carries neither. `scripts/test.sh` skips T3 automatically when no ROM has been imported.

## Verification

Current `dev` is merged into the branch (it had moved 61 commits and added `mapOverview` / map-tile-shading to this same file; `startWildBattle` sits alongside them with no clash).

Full suite on clean `origin/dev` and on this branch, both against freshly regenerated Red data: **17 failures each, failure sets byte-identical.** They trace to a missing `data/generated/audio.lua` — `tools/build_rom_data.py` has no `audio` stage, so any dev-builder dataset fails the audio-dependent suites. Pre-existing and unrelated to this change.

`tests/mod_world_tests.lua`: **19083 → 19100** checks, all passing.

`scripts/test.sh` with `data/generated/` removed (what CI sees): all tiers pass. One caveat on that: `T0 ROM builder version routing` fails locally for me because `scripts/test.sh` invokes bare `python3` and this machine has no Pillow there — it passes when Pillow is on the interpreter, and `ci.yml` installs it.

---

Root-cause analysis and this patch were done with Claude Code; the behaviour was verified on hardware before and after.


crusty:
Specifics on the "written three times over" line above, since it's the part that argues for the API existing.

The mod is [gamecorner-033/Gen1PC-OverworldEncounters](https://github.com/gamecorner-033/Gen1PC-OverworldEncounters). It replaces random grass encounters with visible overworld sprites, so it starts every wild battle itself:

```lua
-- main.lua:383, current release
if battle then Game.stack:push(battle) end
```

No `onFinish`, and no `pushBattle`. Reported as [#14](https://github.com/gamecorner-033/Gen1PC-OverworldEncounters/issues/14) on 2026-08-06 — a Weedle taken to level 10 that never evolved.

Three fixes, none in the shipped release:

- **#4** "Fix post-battle evolution callback" — **merged** 2026-08-03. Its merge commit is no longer in the repository.
- **#15** "Fix evolutions after visible overworld encounters" — closed by its own author.
- **[#17](https://github.com/gamecorner-033/Gen1PC-OverworldEncounters/pull/17)** — mine, so weigh it accordingly. Closed 2026-08-07T20:11:09Z, the same second as a `base_ref_force_pushed` on `main`. A PR can't outlive its merge base being replaced, so it closed automatically.

The repo's history is currently two commits, both "Upload v0.0.6.x", so the merges of #1, #4, #9, #10 and #12 are all gone. That's a workflow problem on their end and squarely their business, not this repo's.

The reason it's worth mentioning here: the bug survived a merged fix, two more attempts, and a full file restructure into Catching/Combat/UI modules — and at no point did anything raise, log, or fail a test. It reads as a working mod right up until you notice a Pokémon that should have evolved. Whatever happens to that mod, an engine-side entry point is what makes this specific mistake unavailable to the next mod that starts a battle.

PR #1063, opened 2026-08-10
Add built-in Spanish (es-ES) interface translation

l0rd92:
This replaces #1032, which GitHub closed automatically after I accidentally
deleted and recreated my fork. The restored branch points to the same commit;
the proposed code is unchanged.

Hi! I’ve been working on a Spanish (Spain) translation for Gen1Recomp’s interface, and I’d like to contribute it to the project.

The idea is to keep English as the default while allowing users to switch between English and Español under Settings. The selected language is applied during startup and remains independent from the language of the ROM or any installed mods.

I’ve also taken the existing translation work into account, particularly @hernan0078’s Spanish mod in #974. This proposal is not intended to replace it: it adds a built-in language option for application-owned text while keeping the existing translation-mod system separate.

## Main changes

- Adds an `AppLocale` layer for application-owned text.
- Loads the saved language once during startup and applies it before the launcher is displayed.
- Updates open screens immediately when the language changes.
- Falls back safely to English if a translation is missing or an unknown language is stored.
- Makes localized save-slot headers reflow responsively instead of shortening translations: the existing one-row layout is preserved when it fits, while longer title/action pairs stack without overlapping the slot list.
- Includes a contribution guide for adding and reviewing more languages in the future.

The translation covers:

- The launcher and Settings.
- ROM and save importing.
- Save-slot management.
- Touch controls.
- Screens used to manage, find, and update mods.
- User-facing notices, confirmations, and errors.

Technical logs remain in English to make diagnostics and error reports easier to search and compare.

## Scope boundaries

This change does not translate or modify:

- ROM content.
- Gameplay dialogue or other game text.
- Save data.
- Names, descriptions, categories, or screens provided by mods.
- The global `Data.strings` catalog or the existing translation-mod system.

Mods can continue to provide their own translated text independently. This proposal also leaves the 8×8 gameplay font unchanged; the application interface uses fonts that support Spanish characters, and accented letters, `ñ`, and punctuation have been checked.

## Validation

- Relevant ROM-free localization and launcher-layout suites pass, including long-string header reflow coverage.
- Manually tested on Windows with LÖVE 11.
- Startup tested from a clean download.
- Immediate switching between English and Spanish verified.
- Language persistence after closing and reopening verified.
- Main screens reviewed at different window sizes.
- The save-slot header was checked at wide and narrow window sizes in both English and Spanish.

Manual validation on Linux, macOS, Android, and iOS is still pending.

This work was prepared by @l0rd92 with AI assistance and manually reviewed on Windows.

Thanks for taking a look. If you would prefer a different approach or a smaller split for easier review, I’m happy to adapt it.


castdrian:
this needs a rebase

PR #1203, opened 2026-08-12
refactor(battle): enhance RNG handling and fixed damage mechanics

1Jamie:
gen2 battles always expose cart battleRandom and gen1/love-style rng using the same stream
effects_static_damage has been properly used and ghost is properly immune where applicable
mangnitude now rolls the right random piece through magnitude_power instead of just self destructing and always returning magnitude 4
added tests to cover the dual rng provisions, sonic boom/dragon rage along with tests for magnitude to prevent regressions


castdrian:
lgtm, however please use feature branches going forward, pr-ing from the default branch is an anti-pattern

castdrian (approved):

1Jamie:
Closing in favor of a feature-branch PR (same commit). Sorry for filing from `dev` — fixed going forward.

Issue #1194, opened 2026-08-12
Emit script.started for onStep events


Yukitty:
### One line summary

Add events/hooks for Prof. Oak escort, Saffron guards, etc.

### Which game is this about

N/A

### What do you want

Add some kind of events or hooks relating to map onStep event scripts, eg. throughout `data/scripts/story2.lua`

### Why is this worth doing

All of the custom onStep map events that stop the player at specific tiles are currently a little difficult for mods to track or edit, because they do not emit the usual "script.start" event, they do not run "world.stepped" or "movement.speed" during overworld scriptMoves, and so on.

### How should it work

- Emit "script.started" or a similar new event when an onStep event meets its trigger condition and begins to take over player control.
- Emit "script.ended" or a similar new event when the previous onStep event has returned overworld control to the player.
- Emit "movement.speed" during scriptMoves for the player, with a new ctx bool marking this is scripted movement?

### Does this change how the original game plays

Quality of life, original game is untouched

### Anything else

Currently, if you use "movement.speed" to give the player running shoes (as a [documented example](https://github.com/bryanthaboi/gen1recomp/wiki/Reference-Hooks#world) on the wiki page for hooks), holding B to run into the grass leaving Palette Town will cause Prof. Oak to stop you, and then forces the player to sprint past him to the lab in half the time, regardless of player input.

This is the most "correct" way I was able to figure out how to resolve that specific issue, and it feels bad because it's VERY indirect?
```
mod.events:on("world.stepped", function(ev)
	if not mod.world:getFlag("EVENT_FOLLOWED_OAK_INTO_LAB") and ev.mapId == "PALLET_TOWN"
	or not mod.world:getFlag("EVENT_BEAT_BROCK") and ev.mapId == "PEWTER_CITY" then
		local ow = mod.world:overworld()
		local player = ow.player
		player.stepFramesCur = player.stepFrames
	end
end)
```

... It is also probably worth considering just resetting the player's stepFramesCur outright for all onStep cutscenes that move the player, lol.

bryanthaboi:
idea noted, closing to clear issue queue, will keep in mind

PR #1210, opened 2026-08-13
fix(OverworldController): ledge jump arc to match player stepFrames

Yukitty:
Adjust ledge jump cosmetic arc for bicycle, running shoes, etc. to still only "hop" the distance of two tiles, not _four_.

Does this match the oracle? I haven't checked, but I would assume so.


Yukitty:
Ah. The alternative may be preferred: Resetting stepFramesCur after every step so that forced steps like ledge falls are always normal speed, even on bicycle. Hm...

That unfortunately makes #1194 related.

PR #1300, opened 2026-08-14
Add manifest-driven required mod imports

anxiousintrovert:
﻿## Summary

- add manifest-driven `required_imports` and `optional_imports` declarations for user-supplied mod files
- validate imports by accepted MD5 values, with N64 byte-order normalization and recognized 512-byte copier-header removal
- store a private canonical copy under each mod's `baseroms/` directory and reuse matching imports from other installed mods
- add an Installed Mods flow to choose, replace, and delete imported files, with validation errors shown directly in the import modal
- block mod loading only when required imports are missing or invalid; optional imports remain non-blocking
- extend Android and iOS picker bridges for generic required imports and prevent ModKit packages from shipping `baseroms/`
- document the manifest format, sandbox boundary, and cross-platform behavior in the modding guide

## Why

Some mods need user-owned source material such as a Stadium ROM, but mod code should not receive arbitrary host filesystem access. This keeps selection and validation in the engine, exposes only the mod's existing scoped `mod:read("baseroms/...")` surface, and avoids distributing copyrighted files.

MD5 is used only as a known-dump identity convention, not as a security primitive.

## User impact

Players manage each declared file from the mod's Installed page. Compatible N64 dumps are canonicalized before validation, a matching file already imported by another mod is copied automatically, and a rejected selection reports its MD5 error in the same modal where the player can choose a replacement.

## Validation

- `luajit tests/mod_required_imports_tests.lua` ? 41/41
- `luajit tests/rom_importer_android_mod_pick_test.lua` ? 26/26
- `luajit tests/launcher_mods_install_zip_test.lua` ? 28/28
- `luajit tests/run_modkit.lua` ? 19/19 suites
- Windows fused payload built and exercised locally

The ROM-backed `tests/run_tests.lua` suite was not available in this checkout because `data/generated/constants.lua` has not been generated.


1Jamie:
seems pretty solid overall! love seeing this come together. there are just a few specific things i wanted to point out before merging.

### big things / security

#### 1. silent cross-mod rom sharing
`RequiredImports.reconcile()` copies a validated file into ANY other installed mod that declared an MD5 that overlaps. it also indexes undeclared files under other mods' `baseroms/`.

a hostile mod wouldn't even need a picker -- it would just need a publicly available md5 and report that's what theirs is. after a player imports let's say stadium 2 for a trusted mod, the other one just gets a private copy automatically and can `mod:read` it. optional imports never even block load.

this makes the "mods never receive arbitrary host files" claim a bit inaccurate. it's not arbitrary host files, but any mod that names a known dump hash can inherit a rom the user imported for someone else.

i'd keep a single engine-owned store and ASK before granting another mod access, or only auto-share with an explicit player toggle / same-author / same-mod-update. if this is intentional i'd declare it and make it known in the docs. if you're trying to stick to the strict "mods never receive arbitrary host files", this would need to be adjusted. not the biggest risk here but worth being honest about.

#### 2. mod-controlled text in shell pickers (this one has to be dealt with)
desktop `chooseRequiredFile` interpolates `spec.name` into `osascript` / `powershell` / `zenity` via `shellSafe`. existing pickers use engine strings, but this one uses mod-authored strings.

the issue is `shellSafe` handles `%`, `"`, and `'`, but it doesn't handle newlines, `$`, backticks, or `zenity`/`osascript` metacharacters. a malicious name becomes an injection surface on linux and mac.

i'd keep a fixed prompt like `"choose required file"` and show the import name only in the love2d modal. i don't think it's wise to put manifest strings into shell command templates -- that's a recipe for g1r to be an attack vector for bad actors.

#### 3. no size cap before read/hash
this one's more of a ux protection. an n64 rom is only tens of MB, but a wrong pick can be GBs. the desktop picker allows `*.*`, which then has the engine read the whole file into a lua string and md5 it. same for the switch MTP inbox (hash every file until one matches) and for `Loader:_validate` on every boot.

a safer and faster bet is an optional manifest `size` / `max_size` and reject over a hard ceiling in general (eg 128mb) before hashing, and cache "this stored file already matched" so boot does not re-hash stadium 2 every launch. that prevents reading files that are already invalid options to start with, prevents hashing things you don't have to, and prevents hashing when the work has already been done once.

inbox matching on switch can still iterate when the player is actually importing. it shouldn't redo that work every time you open the launcher.

---

### a few smaller problems i noticed

* **hidden-file collision:** destination file can be `.something.removed`, which is the same pattern as the removal marker.
* **android pickFile behavior change:** android `pickFile` now returns `false` for unknown kinds instead of falling back to `picked_rom.gb`. safer, but it is a behavior change -- worth a quick one-line note.
* **install preserve failure debris:** an install preserve failure can leave `mods/<id>/baseroms/` with no manifest. discovery should not treat that as a real mod.
* **re-hashing on list refresh:** `LauncherMods.list()` calls `reconcile()` every refresh, which reads and hashes every imported rom for every mod. that will hitch as more things use this feature. if you have 4-5 mods doing this every refresh it will start to build up pretty quick.

---

### quality of life / dev experience recs

a few recs i'd make that i wouldn't say really matter, but might make the dev experience better. up to you on these:

* **import descriptions:** consider adding a `description` (and maybe `hint` / `url`) on each import declaration. *"Pokemon Stadium 2 (USA) .z64, No-Intro"* is what players actually need -- `name` + `file` feels like it might not be enough.
* **mod reading helpers:** expose something like `mod:hasImport("stadium2")` and `mod:readImport("stadium2")` so authors don't have to hardcode `baseroms/stadium2.z64`. that way if they rename a file later, call sites don't break.
* **error context for formats:** if the format is n64, the picker can still be `*.*` which is fine, but the error should say *"expected an n64 rom (.z64/.v64/.n64)"* so users have context when something doesn't work.
* **game-scoped imports:** consider game-scope imports the way dependencies can be game-scoped. a dual-gen mod should not block a red boot on a stadium rom it only needs for gold.
* **iOS picker cleanup:** iOS still has a leftover `stadium` picker kind next to `required_import`. that's fine for compatibility atm, but document which one new code should call so it can be killed off eventually.
* **MD5 vs SHA-1 docs note:** SHA-1 is already the engine's ROM identity, but this feature uses MD5 because ROM DBs do. i'd call this out in the docs so people do not paste SHA-1 into `md5`.

anxiousintrovert:
Addressed the required-import review feedback in `ffa44c2d` and rebased the branch onto the latest `dev`.

Changes made:

- Removed automatic cross-mod ROM discovery/copying. Each import is now an explicit, private player grant for that mod.
- Removed manifest-controlled text from the desktop shell picker prompt; the native prompt is now fixed engine-owned text.
- Added optional manifest `size` and `max_size` fields plus a hard 128 MiB ceiling. Reported file sizes are checked before reading or hashing, including the NX inbox and desktop source path.
- Added validation receipts keyed by accepted digest, file size, and modification time so unchanged imports are not reread/rehashed during launcher refreshes or later loads. Stale receipts are invalidated.
- Changed removal/validation marker names to an engine-owned prefix and reject hidden import filenames, avoiding `.something.removed` collisions.
- Failed baserom preservation during mod replacement now removes the incomplete mod tree and stages recovery under `imports/baseroms-recovery/<mod-id>/`; a retry restores the files and clears recovery.
- Documented Android's unknown picker-kind behavior and the legacy iOS `stadium` picker compatibility path.
- Added optional import descriptions to the launcher UI, clearer N64 format errors, and documentation distinguishing required-import MD5 values from the main importer SHA-1 values.

I left mod-facing helper APIs and game-scoped imports out of this revision since those were optional and would broaden the API/design surface.

Validation:

- required-import tests: 52/52
- Android required/save picker tests: 28/28
- mod install/update preservation tests: 34/34
- modkit: 20/20 suites
- current GitHub Actions checks: passing

1Jamie:
all my qualms are sated here. The RCE was the big concern and you did a great job dealing with it. Also solid fixes on the performance/memory side, and recovery is cleanly handled now too.

1Jamie (commented):
looks good to me, safe too merge

anxiousintrovert:
> all my qualms are sated here. The RCE was the big concern and you did a great job dealing with it. Also solid fixes on the performance/memory side, and recovery is cleanly handled now too.

I will follow up with the other suggested fixes next week, any further requests or suggestions are welcome! I will submit those clean up fixes as a new pr so this can be merged

1Jamie:
> > all my qualms are sated here. The RCE was the big concern and you did a great job dealing with it. Also solid fixes on the performance/memory side, and recovery is cleanly handled now too.
> 
> I will follow up with the other suggested fixes next week, any further requests or suggestions are welcome!

sounds good! i'll let you know if I spot anything else, or if i think of any other DX improvements for mods and pitch them at ya.

bryanthaboi:
Now just fix conflicts

PR #1324, opened 2026-08-15
Fix/save editor lifecycle

1Jamie:
take a look at this in the morning after you sleep, those hooligans can wait.

basically, android wasn't liking how the save editor was switching and accessing stuff across contexts with the new Gold game support added:

Android Editor Crash (Catalog.lua): When opening the save editor on Android, it tried shelling out (io.popen) to scan data/scripts. Android threw a fit because terminal commands are restricted on mobile, returning nil and crashing ipairs on open. Wrapped io.popen in pcall and defaulted to {} so Android stays happy.
Editor State Leaking (main.lua & Data.lua): When swapping between save slots or closing and re-opening, the save editor modules and _pristineKeys stayed cached in package.loaded. Switching between RBY and Gold slots was carrying dirty state across contexts. closeEditor() now wipes the editor modules out of package.loaded and resets Data so every session gets a fresh clean slate.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -166,6 +166,11 @@ function closeEditor()
     require("src.import.CacheFs").unmountVersion(version)
     require("src.core.Data"):unloadGenerated()
   end
+  for k in pairs(package.loaded) do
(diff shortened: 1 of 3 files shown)


1Jamie:
Closing these for right now, I want to check something when I get up first to make sure this is the exact fix and not multiple to get it working right on Android. I noticed a specific behavior I want to look into to make sure I'm matching that pattern after looking closer at it's code

PR #1327, opened 2026-08-15
Render WIDE battle HUD in window space

syybott:
## Summary

Render the existing WIDE battle HUD in physical window space instead of constraining it to the native 304×144 battle surface.

This allows status panels, command boxes, move selection, and message boxes to use the visible window area surrounding the battle scene without changing the existing HUD design or battle logic.

## What changed

- Add a transparent battle HUD layer separate from the native battle composition
- Composite WIDE status panels and battle boxes against the physical window edges
- Keep HUD elements inside the physical window at small window sizes
- Preserve battle shake, palettes, HP bars, text, menus, animations, and input behavior
- Preserve existing menu and overlay stacking
- Keep classic battle layout behavior unchanged

## Root cause

The WIDE battle scene and Gen 1 HUD were flattened into the same 304×144 canvas before that canvas was scaled into the application window. Once flattened, the compositor could not position HUD regions beyond the native battle surface.

The HUD is now drawn into a separate transparent native-pixel layer and placed during the final window composite.

## Compatibility

The new window-space HUD path is limited to `BATTLE LAYOUT = WIDE`.

Classic battle layouts and normal screen-space UI anchoring remain unchanged. The implementation was also tested with Dramaless Shape and Stadium Battle FX enabled.

## Validation

- `battle_fixed_menu_scale.lua`: 32/32
- `ui_layout_option.lua`: 20/20
- `wide_battle_layout.lua`: 33/33
- `wide_battle_shake_bug562.lua`: 40/40
- `battle_fit_option.lua`: 28/28
- Total: 153/153 focused checks passed
- Repeatable fullscreen screenshots captured using `POKEPORT_DRIVER`
- Automated screenshot configuration uses `BATTLE SIZE = FILL`
- FIXED screenshots captured manually with Dramaless Shape and Stadium Battle FX

#### Before: HUD constrained to the native battle composition

Before: HUD constrained to the native battle composition

#### After: HUD positioned in physical window space

After: BATTLE SIZE = FILL

#### Before: HUD constrained to the native battle composition

Before: BATTLE SIZE = FIXED with Dramaless Shape and Stadium Battle FX

#### After: HUD positioned in physical window space

After: BATTLE SIZE = FIXED with Dramaless Shape and Stadium Battle FX

### DIFF
--- a/src/battle/WideBattle.lua
+++ b/src/battle/WideBattle.lua
@@ -92,6 +92,40 @@ local function levelAt(battle, battler, x, y)
   end
 end
 
+-- Move an already-rendered boxed HUD region out of the native 304x144 battle
+-- surface during the physical-window composite.  `windowHUD = false` is a
+-- deliberately battle-local escape hatch for screenshot A/B drivers; normal
+-- WIDE battles use the window layout.  The ordinary UI anchor gate remains
+-- held, so battle-owned TextBox / ChoiceBox states do not move with these.
+local function battleIsTopState(battle)
+  local stack = battle.game and battle.game.stack
+  return not (stack and stack.top) or stack:top() == battle
+end
+
+local function anchorHUD(battle, x, y, w, h, anchor)
+  if battle.windowHUD == false then return end
+  -- Full-screen battle overlays own the UI above the still-running arena.
+  -- Since the detached HUD canvas is normally composited last, registering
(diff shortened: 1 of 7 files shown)


bryanthaboi:
Maybe as its own option?

syybott:
> Maybe as its own option?

yeah I think Ive got a much better implementation cooking now, will update my PR soon

PR #1434, opened 2026-08-16
Eight engine fixes found while building mods against v0.1.98

sanjinpepic:
Eight fixes and one comment correction, found while building mods against v0.1.98. Each commit is self-contained and independently revertable, so any of them can be dropped without disturbing the rest.

Every fix was checked by reverting the source change and confirming the new test fails the way the original symptom describes, then reapplying. Five test files are new; more than half the added lines are tests.

## The fixes

**`drainHold` is never returned to nil.** `stepHPDrain` counts it down but lands on `0` rather than `nil`, and `BattleSafety.inspect` uses `drainHold ~= nil` as part of its "presentation has settled" gate. So the first HP change of a battle leaves it permanently at `0` and every later checkpoint attempt is refused with `battle_phase_busy`, long after the bar has caught up. Now cleared only once the pixel position, the shown HP and the `draining` flag have all settled.

**`require("jit.util")` walks through the mod sandbox.** `DENIED_PREFIX` covers `love.*` and `ffi.*` but not `jit`, and `jit.util` is LuaJIT's own equivalent of the `debug` library the sandbox already denies by name — `funcbc` and `funck` read the bytecode and constants of any reachable function, which is enough to recover the upvalues (`_G`, `love`, `io`) the sandbox exists to withhold. Adds `jit` to the same prefix list, leaving the bare `jit` global and `require("jit")` untouched.

**A TM or HM used on a species with no `tmhm` list crashes.** `ItemEffects` calls `ipairs(speciesDef.tmhm)` unguarded, so any species lacking the field takes the game down instead of refusing. Now refuses cleanly.

**`Game2:usePartyItem` drops its `data` argument** when calling `ItemEffects.partyAction(itemId)`, so the lookup only ever sees the engine's built-in item table and never a merged one — no registered Gen 2 field item is reachable through USE at all. The identical omission is fixed at a second call site in `BattleState:useItem`.

**`Mon.syncIdentity` recomputes `mon.shiny` unconditionally.** It runs from `refreshStats`, which `SummaryMenu.new` calls on every menu open, so a forced shiny is cleared whenever its DVs do not match the natural pattern — opening the summary screen un-shinies a scripted shiny. The recompute is now monotonic.

**The save editor's Gold bootstrap omits `gen2Constants`.** `Gen.lua`'s `bindGoldData` binds `gen2Maps`, `gen2Tilesets`, `gen2Palettes` and others but not that one, so anything reading `constants:get("generation")` misreads the generation under the editor and shapes its records for the wrong game. In our case 1238 species registrations were silently refused while the same mod worked correctly in-game.

**A map object's `pokemon` field is never validated.** `R.maps.objects` is `f.opt(f.list(f.any))`, so unlike an encounter slot a bad species id in a map object survives load and becomes a crash when the player walks up to it. Validated now, which needed a small `f.partial` schema primitive — flagged below as the most invasive change here.

**An `item.use` hook around `BagMenu`'s dispatch.** `useOn` is a file-local, so there is no seam anywhere around item use; every other screen a mod can reach is a table field. This adds a `Runtime.call` hook rather than exporting the local, as the smaller of the two options — but it is an API addition and therefore your call on shape. Happy to drop this commit if you would rather it looked different.

## One report that turned out wrong

`Schemas.GEN1` not gating `growth_rates` / `evolution_methods` looked like the cause of a large number of spurious unresolved-reference errors under a Gen 2 loader, and the comment above `crossValidate` describes exactly that gating. It is not a bug: implementing it would break `gate_gen2_mod_api.lua`, which deliberately asserts those registries are not gated, and the errors trace to a data gap in a headless harness instead. Only the comment is changed, since it is what led us to the wrong conclusion in the first place.

## Notes

Branched from `v0.1.98` and rebased onto current `dev`; the new tests pass on that base. The remaining failures in the suite here are pre-existing and environment-specific (Windows path separators, headless shell restarts) and unrelated to these changes.

Happy to split this into separate PRs, reshape anything, or drop individual commits.

### DIFF
--- a/src/battle/BattleState.lua
+++ b/src/battle/BattleState.lua
@@ -1103,6 +1103,15 @@ function BattleState:stepHPDrain()
       if not b.shownPx then b.shownPx = targetPx end
       if (b.drainHold or 0) > 0 then
         b.drainHold = b.drainHold - 1
+        -- Once the count runs out with nothing left pending (bar and
+        -- number already on the final total), the drain is over, not just
+        -- between steps: leave the field at 0 and BattleSafety.inspect
+        -- reads it as still mid-animation for the rest of the battle,
+        -- since drainHold ~= nil is its settled-presentation gate.
+        if b.drainHold <= 0 and b.shownPx == targetPx and b.shownHP == goal
+            and not b.draining then
+          b.drainHold = nil
+        end
         busy = true
       elseif b.shownPx ~= targetPx then
         -- .barAnimationLoop redraws the bar one pixel at a time, `ld c, 2 /
--- a/src/battle/gen2/Mon.lua
+++ b/src/battle/gen2/Mon.lua
@@ -112,7 +112,13 @@ function Mon.syncIdentity(mon, data)
   if mon.dvs then
     mon.gender = Mon.gender(def, mon.dvs,
       { species = mon.species, level = mon.level })
-    mon.shiny = Mon.isShiny(mon.dvs,
+    -- shiny is monotonic once true, the same as opts.shiny winning over
+    -- shiny.roll at Mon.new: a forced shiny (the scripted-shiny path, DVs
+    -- that do not themselves read as shiny) must not un-shiny the moment
+    -- this runs again, and it runs on every SummaryMenu open via
+    -- refreshStats.  A mon not already shiny still promotes normally if
+    -- its DVs justify it, e.g. after an edit.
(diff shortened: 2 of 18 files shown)


bryanthaboi:
Tests fail

bryanthaboi:
Update ur branch might fix the tests failing

sanjinpepic:
I think you are correct regarding the update of branch

Issue #597, opened 2026-08-01
Pulling mod index fails on Android


dejbob:
### One line summary

Red error message „curl is not available on this platform“ when adding a mod index.

### Which game were you playing

Red

### Which build are you running

Android

### Were any mods on

No

### Version

0.1.52

### Where in the game

Loader

### Screenshot, or a detailed description if you couldn't take one

<img width="3024" height="4032" alt="Image" src="https://github.com/user-attachments/assets/c6ffdbb7-c4a4-4ba3-9669-e4f553cdd453" />

### How to make it happen

Just put in the url or author/repo of the mod index github

### What should have happened

I guess i should see the mods

PR #1436, opened 2026-08-16
feat(android): add httpPost bridge for mod.postLog log sends

ShaneMcGovernIE:
## Problem

`mod.postLog` fails on Android with **"no POST transport on this platform"** (HostShell.lua). Android ships no curl binary and the JNI bridge (`love.system.httpDownload`) is GET-only, so the only POST path a log send has on desktop is absent.

Observed on-device (Android, engine 0.1.98, mod 1.7.1): every manual send and every auto-send attempt fails with that error, and the failure line is force-persisted into the mod's storage each interval.

## Change

Adds the mirror of the existing `httpDownload` transport, end to end:

| File | Change |
|---|---|
| `GameActivity.java` | `httpPost(url, byte[] body, contentType, userAgent)`: https-only, hand-followed redirects **re-POSTing the body on each hop** (curl `-X POST` behaviour), one-way — response drained and discarded, 2xx verdict only. Runs on the Lua/worker thread, blocking, like `httpDownload`. |
| `common/android.cpp` / `.h` | JNI bridge `httpPost`, same pattern as `httpDownload`: activity class resolved via `SDL_AndroidGetActivity()` (never `FindClass` — worker threads), old-APK skew tolerated by null-checking the method id (#597). Body crosses as a **jbyteArray** so arbitrary UTF-8 in a log ring cannot be corrupted by modified-UTF-8 jstring conversion. |
| `System.cpp` / `.h` | `System::httpPost`, Android-only, `LOVE_UNUSED` elsewhere. |
| `wrap_System.cpp` | `love.system.httpPost` binding + method table entry. |
| `HostShell.lua` | `httpPost` uses the bridge when curl is absent; keeps the "no POST transport" fallthrough for builds without it (old-APK skew reads as a missing transport, never a crash). |

No changes to the mod sandbox or `mod.postLog` API — `fetch_worker.lua` → `Net.postLog` → `HostShell.httpPost` picks the bridge up automatically.

## Notes

- iOS is still bridge-only without a POST arm; a URLSession sibling will follow in a later PR.
- Tested from the mod side against a loghook server: the failure line disappears and log posts land.

### DIFF
--- a/mobile/android/love/src/jni/love/src/common/android.cpp
+++ b/mobile/android/love/src/jni/love/src/common/android.cpp
@@ -325,6 +325,55 @@ bool httpDownload(const char *url, const char *destPath, const char *userAgent, 
 	return result;
 }
 
+bool httpPost(const char *url, const char *body, int bodyLen, const char *contentType, const char *userAgent)
+{
+	if (url == nullptr || body == nullptr || bodyLen < 0)
+		return false;
+
+	JNIEnv *env = (JNIEnv*) SDL_AndroidGetJNIEnv();
+	// Same resolution rule as httpDownload: the activity's own class via
+	// SDL_AndroidGetActivity, never FindClass -- this bridge is called off
+	// the main thread (love.thread workers), whose class loader cannot see
+	// app classes.
+	jobject activityObj = (jobject) SDL_AndroidGetActivity();
+	if (activityObj == nullptr)
+		return false;
+	jclass activity = env->GetObjectClass(activityObj);
+	env->DeleteLocalRef(activityObj);
+
+	// Old APK / new liblove skew: report "no transport" the same way a
+	// missing curl does, instead of aborting on a missing method (#597).
+	jmethodID method = env->GetStaticMethodID(activity, "httpPost",
+		"(Ljava/lang/String;[BLjava/lang/String;Ljava/lang/String;)Z");
+	if (method == nullptr)
+	{
+		env->ExceptionClear();
+		env->DeleteLocalRef(activity);
+		return false;
+	}
+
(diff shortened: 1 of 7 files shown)


ShaneMcGovernIE:
**Investigating the headless-suite failure: it's pre-existing on `dev`, not from this PR.**

The failing tests in the CI run are all gen2 suites (`gen2_fishing_bite_gate_bug1368`, `gen2_grass_encounter_tod_bug1389`, `gen2_map_bake_dpi`, `gen2_render_output_seam`, `gate_gen2_mod_facade`, …) failing with `error loading module 'src.world.gen2.World'` and fixture-reference errors. Evidence this isn't this PR:

- `dev`'s own CI is already red on the same tier: runs [31957769611](https://github.com/bryanthaboi/gen1recomp/actions/runs/31957769611) and [31957723150](https://github.com/bryanthaboi/gen1recomp/actions/runs/31957723150) (recent dev pushes, both `failure`).
- Running `./scripts/test.sh` on this branch locally: every gen2 suite passes; the only failing suites are the two `mods/potato_voxel` ones, which is the known `POKEPORT_DATA_DIR` gap (`tests/run_modkit.lua` runs mod suites without it; both pass 262/262 with `POKEPORT_DATA_DIR=tests/fixture_data`). CI never mounts mods, so it can't see those anyway.
- The suite that exercises this PR's code — `host shell fetch errors` — passes (21/21).

The patch touches only `HostShell.lua` and the Android bridge files; nothing gen2-related. Happy to rebase if a clean red-free baseline is wanted once `dev`'s gen2 breakage is fixed upstream.

bryanthaboi:
Update ur branch

PR #1480, opened 2026-08-17
feat: support multiple connections per map side

Vuelos:
Allow a map side (north/south/east/west) to carry an array of connections instead of a single connection object. The engine's connection landing and neighbor computation now iterate all connections on a side, picking the one whose seam sits under the player's position along the edge.

Changes:
- Map.connection() may return a connection object or an array of connections
- OverworldState.computeNeighbors iterates all connections on each side
- World.computeNeighbors (gen2) does the same
- connectionLanding resolves the correct connection by player cell position
- Gen2Compat.connectionLanding uses the same multi-connection picker
- New OverworldState:connectionAt() helper for edge-exit picking

A side that carries only one connection continues to behave exactly as before.


bryanthaboi:
Example of use case?

castdrian:
please rebase your branch instead of backwards merging from upstream, that will avoid polluting the commit history

Vuelos:
I'm using this on my mod to add multiple connections walkable to different maps at the side of long maps like ROUTE 17 on gen 1.
Being able to connect PALLET TOWN, VIRIDIAN CITY and ROUTE 1 to ROUTE 17 with small maps filling the space.

The renderer and other changes are for using any tileset on any map without issues with the pallete.

I'll like those things available on the base recomps but I'm not sure if this is the best approach as it is heavily vivecoded.

Marked as draft.

PR #1559, opened 2026-08-19
Route more battle/overworld/menu messages through their real ROM text

thibautbus:
## Bug

Twenty-one real ROM labels, across call sites in `BattleState.lua`, `OverworldController.lua`, `BoxMenu.lua`, `SlotMachine.lua`, `ShopMenu.lua` and `LinkBattle.lua`, called `Strings()` with an engine-authored literal instead of the real, already-extracted ROM text label -- in several cases with a comment right next to the call already naming the real label. `romText()`/`self:romText()` is the established pattern for this throughout the codebase (dozens of existing call sites); these were simply missed.

Fixing them has two genuinely different effects depending on the site, so this PR splits them below rather than treating them all the same way.

### Real translation changes

For these, the engine literal was a documented translation compromise (an adapted/approximate stand-in, or one literal shared across several incompatible ROM contexts) -- routing through the real label changes what actually renders:

- **`storeCaughtMon()`'s box-transfer message**: previously one `Strings()` key ("%s was\ntransferred to\n%s!") with a hardcoded "BILL's PC"/"someone's PC" swapped in as if it were a substituted argument. `_ItemUseBallText07`/`_ItemUseBallText08` are two full, independently worded ROM strings, not one template with a swapped name -- picked by `EVENT_MET_BILL` instead.
- **`throwBall()`'s dodge + can't-be-caught lines**: two separate `Strings()` calls, each independently translated, merged into `_ItemUseBallText00` (one `\f`-paged ROM label covering both). Unlike `TextBox.new()`, which splits `\f` itself, `sayNext()` goes through the battle queue's own `startMessage()`, which only splits on `\n`/`\v` -- confirmed live in a real build (the second sentence overflowed off the box instead of starting a fresh page). Fixed by resolving the label once and splitting it the same way `TextBox.lua` does before queuing one `sayNext()` per page.
- **`onFaint()`'s faint message** and **`applyFieldPoison()`'s field-faint message**: previously one shared `Strings("%s\nfainted!", ...)` key covering three different original ROM contexts (enemy, player, field). `_EnemyMonFaintedText` already carries its own "Enemy " wording in the ROM text itself, so the old shared key's separate `Strings("Enemy %s", ...)` composition doubled it up if reused naively; now picks `_PlayerMonFaintedText`/`_EnemyMonFaintedText`/`_PokemonFaintedText` per context, passing the raw name.
- **`useSoftboiledFieldMove()`'s no-effect and heal messages**: `_ItemUseNoEffectText` retires this call site's share of a literal shared with `ItemEffects.lua`'s own item failures. `_PotionText` is the more notable one: the old literal ("%s's HP\nwas restored!") never had a slot for the amount healed at all -- `_PotionText`'s real text has a second slot for it (`{NUM:wHPBarHPDifference}`), the same label and shape `ItemEffects.lua`'s own potion-item message already uses. Verified against `pret/pokered`: Softboiled/Milk Drink used as a field move sets `wPartyMenuTypeOrMessageID` to the same `POTION_MSG` regular Potion items use, so this is the correct shared ROM label, not a coincidental reuse.
- **`tryHiddenObject()`'s hidden-item finds** and the **normal item-ball pickup path**: `_FoundHiddenItemText`/`_FoundItemText` retire a literal ("%s found\n%s!") shared across both the hidden-object and normal-pickup ROM contexts.
- **`BoxMenu.lua`'s RELEASE confirmation and follow-up**: `_OnceReleasedText` retires an adapted-argument-order compromise. `_MonWasReleasedText`, the "X was released outside. Bye X!" message shown right after confirming, was missed in an earlier pass of this PR and only caught testing the flow live -- same `t._X or Strings(...)` pattern as the line above it.
- **`SlotMachine.lua`'s lined-up message**: `_LinedUpText` retires a compromise that included the slot symbol as if it were part of the translatable sentence, when the ROM text never had a slot for it at all (the symbol is drawn separately; concatenated in front of the label now).

### Robustness only, no visible change

For these, the pre-fix `Strings()` literal is byte-identical to a real ROM text fragment, so `pipeline/engine.py`'s exact corpus alignment already matches and translates it automatically today, no manual override involved. Routing these through their real label changes nothing a player sees today; the value is resilience against a future wording change to the engine literal silently breaking the coincidental exact-match (which would revert to English with no visible failure until someone notices):

- **`enter()`'s pre-battle black-out message**: two separate `Strings()` calls joined with `\f`, merged into one `_PlayerBlackedOutText2` call.
- **The AI switch-in withdraw/send-out line and the enemy trainer's first send-out** (3 call sites, one shared by the link-battle intro path): `_AIBattleWithdrawText`/`_TrainerSentOutText`.
- **`ShopMenu.lua`'s buy/sell price confirmations**: `_PokemartTellBuyPriceText`/`_PokemartTellSellPriceText`. This file's own `txt(game, key, fallback)` helper doesn't support substitution arguments, so it can't be reused as-is -- added the module-level `romText` helper instead, same as every other file in this batch.
- **`LinkBattle.lua`'s opening line**: `_TrainerWantsToFightText`. The battle object built at this point is already a `BattleState`, so this reuses its `self:romText` convenience method.

### Investigated, not changed

- **`_TrainerAboutToUseText`** (the SHIFT-switch prompt, `enemyMonFainted()`): tried merging its `say()` + `sayChoice()` pair into one `\f`-paged `romText()` + `sayChoice()` call, the same way as `_ItemUseBallText00` above. `tests/engine/trainer_shift_prompt_bug565.lua` caught it immediately (wrong line count, no `\f` paging) -- the battle queue's own text renderer behind `sayChoice()` doesn't page `\f` the way `TextBox.lua` does for `sayNext()`/`say()`. Left as the original two separate `Strings()` calls, with a comment recording why.
- **`MoveEffects.lua`'s stat-rise/fall messages**: already routed through `Strings()` with a shared, already-translated template (`"%s's %s\nrose!"`); not a `romText()` candidate, no change needed.

## Testing

- Seven new tests in `tests/engine/` (one per message family), each driving the real interactive flow or faking `Game`/`TextBox` via `debug.setupvalue` where there's no state to drive directly, asserting the routed text against a faked real label plus a vanilla no-catalog fallback case: `box_release_confirmation_romtext.lua`, `slot_machine_lined_up_romtext.lua`, `battle_fainted_message_romtext.lua`, `battle_catch_messages_romtext.lua`, `battle_ball_dodge_romtext.lua`, `overworld_field_faint_heal_romtext.lua`, `overworld_hidden_item_romtext.lua`. Not every fixed call site has its own dedicated test -- `ShopMenu`, `LinkBattle`, the trainer withdraw/send-out sites and the normal (non-hidden) found-item sites share the same `romText` mechanism already proven correct by the seven tests above.
- Every new `romText`/`self:romText` call's fallback format string and argument count/order cross-checked against the real extracted ROM label in `data/generated/text.lua` (token counts and ordering, including the `{PLAYER}`-before-item-name ordering `overworld_hidden_item_romtext.lua` specifically guards against).
- `luajit tests/run_engine.lua`: 236/236 suites pass.
- Manually verified in-game with the French translation mod loaded (Windows build), comparing `dev` against this branch side by side for 9 of the 11 distinct scenarios (the slot machine's 3-symbol match and a link battle were left to the automated/static checks -- both awkward to trigger on demand). Every "real translation change" scenario showed the expected new wording, most notably `_PotionText` now showing the actual HP amount recovered, which the old literal never displayed at all. This pass also caught the `_ItemUseBallText00` `\f`-pagination bug described above, which none of the automated tests had coverage for.

### DIFF
--- a/src/battle/BattleState.lua
+++ b/src/battle/BattleState.lua
@@ -1718,9 +1718,14 @@ function BattleState:enter()
     -- _PlayerBlackedOutText2 (data/text/text_2.asm:896): the two paragraphs
     -- playerMonFainted queues on the battle screen; there is no battle
     -- screen to queue them on here, so they print over the map.
+    -- _PlayerBlackedOutText (no "2") extracts to the identical wording from
+    -- a different ROM address and is unused anywhere in this engine -- not
+    -- a fallback for this one, just pokered printing the same paragraph
+    -- from a second call site elsewhere.
     self.game.stack:push(require("src.render.TextBox").new(self.game,
-      Strings("%s is out of\nuseable POKéMON!", name) .. "\f"
-      .. Strings("%s blacked\nout!", name), blackedOut))
+      self:romText("_PlayerBlackedOutText2",
+        "%s is out of\nuseable POKéMON!\f%s blacked\nout!", name, name),
+      blackedOut))
     return
   end
   self.musicKind = self:computeMusicKind()
@@ -1857,7 +1862,8 @@ function BattleState:enter()
       self:slidePic("foe")
     end)
     -- _TrainerSentOutText ends `done`, not `prompt` (data/text/text_2.asm:923)
-    self:sayAuto(Strings("%s sent\nout %s!", foeName, self.enemy.name))
+    self:sayAuto(self:romText("_TrainerSentOutText", "%s sent\nout %s!",
+      foeName, self.enemy.name))
     self:act(function()
       -- EnemySendOutFirstMon (core.asm:1421-1434): after the text the
       -- pic grows out of the ball (AnimateSendingOutMon), then the cry
@@ -3638,12 +3644,13 @@ function BattleState:executeAction(user, target, action)
       })
       self.aiUses = self:aiUsesFor()
       markSeen(self.game, self.enemy.mon.species)
-      -- _AIBattleWithdrawText: "X with-/drew Y!"
-      self:sayNext(Strings("%s with-\ndrew %s!", self.trainer.name, oldName))
+      self:sayNext(self:romText("_AIBattleWithdrawText", "%s with-\ndrew %s!",
+        self.trainer.name, oldName))
       -- EnemySendOut falls into EnemySendOutFirstMon: TrainerSentOutText,
       -- then AnimateSendingOutMon and PlayCry (core.asm:1276-1434)
       self.enemySendingOut = true
-      self:sayNextAuto(Strings("%s sent\nout %s!", self.trainer.name, self.enemy.name))
+      self:sayNextAuto(self:romText("_TrainerSentOutText", "%s sent\nout %s!",
+        self.trainer.name, self.enemy.name))
       self:actNext(function()
         self.enemySendingOut = false
         self:startGrowIn(self.enemy)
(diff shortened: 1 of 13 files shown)


bryanthaboi:
merge conflicts

thibautbus:
@bryanthaboi the branch has been rebased

PR #1571, opened 2026-08-19
feat(macOS): default to löve12 and metal as gpu backend

castdrian:
- merging this will require things such as voxel mods to adjust their rendering for the löve12 behaviour, most notably the y positioning


bryanthaboi:
if we are gonna update to love 12 we need to update all not just mac

castdrian:
> if we are gonna update to love 12 we need to update all not just mac

true but I can't really test binbows/linux/androod builds so I figured this was safest for now analog to ios, but I can swap the rest as well :suminidktbhprobablynod:

bryanthaboi:
we can revisit this at a later date

Issue #1575, opened 2026-08-19
Modkit's validator incorrectly flags assets as derivative of boulder.png


dburton95:
### One line summary

Modkit's validator incorrectly flags assets as derivative of boulder.png

### Which game were you playing

N/A

### Which build are you running

Linux

### Mods off

- [x] I turned off all mods and can still reproduce this

### Version

0.0.0-dev

### Discord username (optional)

dgray66

### Where in the game

Not in game, in modkit.py

### Screenshot, or a detailed description if you couldn't take one

The validate function in modkit.py flags any 16x16 opaque or transparent asset as a derivative of boulder.png.

### How to make it happen

Make a 16x16 sprite thats opaque or just a 16x16 image with nothing in it and run the modkit's validate tool.

### What should have happened

Shouldn't have flagged these as stolen assets

### Anything else

I have a fix for this already. When the validator downscales boulder.png, it's pixels are so opaque and dense that pretty much anything thats 16x16 and a little opaque is flagged. Changing the greyscale threshold from a fixed integer to an average of the image's gradient resolves this. I'll post the PR next.

dburton95:
Fix posted here #1576

PR #1576, opened 2026-08-19
Fixes the greyscale cutoff for boulder.png

dburton95:
A fixed cutoff (e.g. "<=200 is ink") only makes sense for sprites with a light background to split against; a mostly-opaque 16x16 icon has almost no pixel above that cutoff, so every such icon collapsed onto the same "all ink" hash and was flagged as a near-duplicate of anything else that also collapsed -- which was most of them, boulder.png included. Thresholding against the image's own mean keeps the split meaningful (and roughly balanced) no matter how light or dark the source is.


dburton95:
fixes #1575

PR #1616, opened 2026-08-20
Fix/large required imports v2

HighDrexler:


1Jamie:
@HighDrexler  nothing crazy but a few things to clean up:

1. potential State Corruption in RomImporter.lua
In streamRequiredImport the global `CacheFs.prefix is mutated to allow writing to the mods/ directory instead of the cache directory.

```lua
local savedPrefix = CacheFs.prefix
CacheFs.prefix = ""
-- ... streaming logic ...
CacheFs.prefix = savedPrefix
```
the problem is if a Lua error occurs during the streeming loop (e.g. inside md5:update or output:write), savedPrefix is never restored. this WILL corrupt ALL subsequent CacheFs operations for the entire session. i would suggesting wrap the streaming logic in a pcall or xpcall to ensure CacheFs.prefix = savedPrefix is always executed in a finally-style block.

i would recomend making the same change to RequiredImports.acceptStoredDigest also it has a similar savedPrefix pattern (lines 228-232) its less likely to cause and issue due to not being a loop, but not having it able to at all is of course always prefered.

2. little bit of instability in openImportSource
the native:seek logic for host file handles could be safer:
```lua
local size = native:seek("end")
if size then native:seek("set", 0) end
return { ... }
```
in this case if seek("end") succeeds but seek("set", 0) fails, the handle is returned with the file pointer at the end. any subsequent reads will return nil immediately, which might be confusing. I would Verify that seek("set", 0) succeeds before returning the handle.

other than this i think its pretty solid, should solve the issue with larger imports and not making the engine crap its pants at the size.

after those changes it should be good to merge

HighDrexler:
* streamRequiredImport now uses guaranteed/finally-style cleanup, so a thrown hash/write error cannot leave CacheFs.prefix corrupted.
* RequiredImports.acceptStoredDigest now has the same protection.
* The native ISO handle now verifies that seek("set", 0) succeeds before it can be returned.
* Regression tests were added for all three failure cases.
*Removed bad text assertion regarding expected rom
*Fixed large required imports needing to be streamed/chunked instead of loaded into one Lua string.

PR #1627, opened 2026-08-20
fix(android): bootstrap legacy APK updater

AverageConsumer:
## Summary

- detect legacy Android shells that lack `love.system.installApk` before the engine-version early return
- route only those shells through the existing full-package requirement and link the exact release APK for one manual bootstrap update
- preserve the current `.love`, `minShell`, and `payloadHost` update behavior for every capable Android shell and every other platform

## Why

An older APK can already be running the latest downloaded `.love`, so its engine version reports up to date even though its native shell predates the APK installer bridge. The existing version early return then clears the full-update requirement and leaves that mixed Lua/native installation stuck.

The old APK cannot invoke Android's Package Installer itself, so this case gets a clear **Update app manually** footer action that opens the matching APK asset. Once that APK is installed over the existing app, the existing selective updater handles future releases: ordinary Lua/data updates stay `.love` updates, while `minShell` or `payloadHost` incompatibilities request a full package.

This does **not** force an APK for every Android update.

## Tests

- `luajit tests/engine/update_check_tests.lua` (33/33)
- `luajit tests/engine/update_tests.lua` (119/119)

### DIFF
--- a/docs/updater.md
+++ b/docs/updater.md
@@ -126,9 +126,12 @@ bundled game, in that case.
    persistent launcher control. Android downloads the release APK, verifies
    its SHA-256 entry from `sha256sums.txt`, then invokes Android's Package
    Installer. The installer asks the user for consent and enforces package,
-   version-code, and signing-certificate compatibility. iOS links the
-   sideload repository for a re-sideload; Xbox, desktop, and PortMaster builds
-   link their correctly named full package. Switch keeps its native OTA flow.
+   version-code, and signing-certificate compatibility. A legacy APK without
+   the installer bridge links its full package for one manual bootstrap
+   update, including when its downloaded payload already reports the latest
+   engine version. iOS links the sideload repository for a re-sideload; Xbox,
+   desktop, and PortMaster builds link their correctly named full package.
+   Switch keeps its native OTA flow.
 
 ## Known limitations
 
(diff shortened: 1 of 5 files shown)


bryanthaboi:
can u review this for what i have already set up - it SHOULD already do the apk situation IF there is need, i.e. the shell updated and not just the .love --- perhaps u can rewrite to maybe enahnce what i already did? not every update actually NEEDS a new apk

AverageConsumer:
True. The only edge case worth keeping is the legacy bootstrap case: an old APK may already report the latest engine version after installing the .love, while still lacking the native installer bridge. I’ll rewrite the PR to only detect that situation and request one full APK update. After that, normal .love updates and your existing shell checks remain unchanged.

AverageConsumer:
Yep, you were right. I rewrote this to preserve the selective updater you already built. It now only catches legacy Android APKs that lack love.system.installApk, including the mixed state where the downloaded .love already reports the latest engine version. Current Android shells still use normal .love updates unless the existing minShell / payloadHost gate requires a full package. All checks are green.

AverageConsumer:
Tested on my AYN Thor with two isolated Android builds.  
A current host with the native installer bridge kept the normal selective update path: it updated from 0.2.16 to 0.2.18 via the .love payload without opening Android’s Package Installer.  
I also built a clean legacy-host simulation without love.system.installApk, cleared all app data, and stamped both the APK and embedded engine as 0.2.18 to match the latest release exactly. On first launch it immediately showed Update app manually, and the action resolved to the correct official 0.2.18 Android APK.

PR #1645, opened 2026-08-21
feat(mod-api): expose charge decision hook

MaxTomahawk:
## Summary

- add the guarded shared `battle.charge_required` hook at the initial-use charge decision in Gen 1 and Gen 2
- let a public mod return exactly `false` to skip only charge setup while preserving the engine-owned PP, accuracy, damage, animation, and effect pipeline
- document the additive API and shared-generation compatibility contract in RFC 0011

## Motivation

Battle-mechanics mods cannot currently skip an existing charge stage without mutating private battler state or replacing the move pipeline. A weather ruleset is one consumer, but the seam is generic and contains no weather, move-id, trainer, or Adaptive Trainers policy.

## Compatibility

This is additive. Existing mods require no migration. The call sites are guarded with `Runtime.wantsHook`, so a no-mod battle allocates no hook context and retains vanilla charge, PP, text, animation, and native Gold sun behavior. Gold applies its existing native rules first; Solarbeam in native sun does not invoke the hook.

## Verification

- `luajit tests/engine/battle_charge_required.lua` — 47/47
- `luajit tests/engine/gate_hooks.lua` — 493/493
- `luajit tests/engine/gate_gen2_mod_api.lua` — 943/943
- `./scripts/test.sh` — all 282 engine suites, all 23 mod-SDK suites, and title checkpoint cold restart passed; ROM-dependent T3 skipped because this clean worktree has no generated cartridge data
- CI-equivalent `modkit lint` loop — passed
- `python3 tools/modkit.py docs` — generator completed; unrelated pre-existing generated-doc drift was not included

## Route B checklist

- [x] RFC with traceable design decision and plan
- [x] backward-compatibility statement
- [x] no-mod parity test with guarded hot path
- [x] public mod-API test through the sandboxed SDK
- [x] shared Gen 1 / Gen 2 catalog gate and compatibility docs
- [x] no deprecations or removals

### DIFF
--- a/CONTRIBUTING-mods.md
+++ b/CONTRIBUTING-mods.md
@@ -173,7 +173,7 @@ before the key existed changes behavior; list both generations or say `"all"`
 when you mean everywhere.
 
 `docs/mod-api-gen2-compat.md` is the compatibility matrix: what works on Gold
-and Silver today (40 of the 46 registries, 40 event and 43 hook names shared with Gen 1,
+and Silver today (40 of the 46 registries, 40 event and 44 hook names shared with Gen 1,
 and 24 Gen 2-only ones), which registries have no Gen 2 home and drop their
 writes with a report, and which hooks and events are still to come.
 `docs/preparing-your-mod-for-gen2.md` is the step-by-step migration guide for a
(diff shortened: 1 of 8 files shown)


bryanthaboi:
@MaxTomahawk from now on, put in ur PR's what made u want to add this feature. as in like what in a mod ur working on, needed this. thanks.

MaxTomahawk:
Understood. Future engine PR descriptions will name the concrete consuming mod, the exact blocked feature, and why the existing public seams are insufficient. I have also made that a binding rule in Adaptive Trainers' AGENTS.md.

Issue #1348, opened 2026-08-15
Gen 2 battle text and the Oak intro speech bypass the mod translation catalog


LordSangreal:
## Gen 2 battle text and the Oak intro speech bypass the mod translation catalog

While building a pt-BR translation mod (VersaoDourada, targets Gold) I
found two related gaps where Gold-specific text never reaches a mod's
`strings.lua`/`dialogue.lua` overrides, even though the equivalent Gen 1
text does.

### 1. Several Gen 2 battle messages are hardcoded English, not `S(...)`-wrapped

Unlike most of `src/battle/gen2/Battle.lua` and `src/ui/gen2/BattleState.lua`,
these specific lines build the message by raw concatenation instead of
going through the `Strings` catalog (`src/core/Strings.lua`):

- **"X sent out Y!"** -- `src/battle/gen2/Battle.lua:3031` and `:3881`,
  `src/ui/gen2/BattleState.lua:363`
  (`text = (self.trainer.name or "TRAINER") .. " sent out " .. self:monName(self.enemy) .. "!"`)
- **"X wants to battle!"** -- `src/ui/gen2/BattleState.lua:352`
  (`text = trainerName .. " wants to battle!"`)
- **"Wild X appeared!"** -- `src/ui/gen2/BattleState.lua:346`
  (`text = "Wild " .. self:name(enemy) .. " appeared!"`)
- **"X gained N EXP. Points!"** -- `src/battle/gen2/Battle.lua:3210-3211`
  (`text = self:monName(mon) .. " gained " .. (traded and "a boosted " or "") .. amount .. " EXP. Points!"`)

For comparison, the fork [Gen2Recomped](https://github.com/UNDERdecoded/Gen2Recomped)
unified the Gen 1 and Gen 2 battle code into one `src/battle/BattleState.lua`,
and the same four messages there are built with
`self:sayNext(Strings("%s sent\nout %s!", self.trainer.name, self.enemy.name))`
-- i.e. they're translatable simply by reusing the code path Gen 1 battles
already have, rather than a Gen-2-specific rewrite. Might be a useful
reference for porting the fix here, even if the rest of the two projects'
battle code has diverged since.

### 2. The opening OAK speech (`_OakText1`-`_OakText7`) can never be translated by a mod, for two stacked reasons

**a. The fallback table is built at require time, with the wrong helper.**
`src/ui/gen2/OakSpeech.lua:58-65`:

```lua
local FALLBACKS = {
[...]

PR #1673, opened 2026-08-21
Print the Gen 2 battle lines the cart actually writes

colsonrice:
Five Gen 2 battle messages print something the cart does not say. They are
written by hand in `src/battle/gen2/Battle.lua` rather than taken from
`data/text/battle.asm`, and each has drifted:

| line | prints today | pokegold |
|---|---|---|
| `:1751`, `:2337` | "Hit 3 **time(s)**!" | `PlayerHitTimesText` / `EnemyHitTimesText`: "Hit 3 times!" |
| `:1972` | "All POKéMON hearing the song will faint in three turns!" | `StartPerishText`: "Both #MON will / faint in 3 turns!" |
| `:1238` | "It's super effective!" | `SuperEffectiveText`: "It's super- / effective!" |
| `:1240` | "It's not very effective**...**" | `NotVeryEffectiveText`: "It's not very / effective**…**" |
| `:1394` | "No PP left for this move!" | `BattleText_TheresNoPPLeftForThisMove`: "There's no PP left / for this move!" |

The one a player actually notices is `Hit 3 time(s)!`, which renders that
parenthetical on screen. Gen 2 has no singular form of the line, so there is
nothing to hedge, and Gen 1 in this repo already prints it correctly, at
`src/battle/EffectRegistry.lua:319`:

```lua
battle:sayNext(romText(battle.data, "_HitXTimesText", "Hit %d times!", hits))
```

Perish Song is the other one worth a look: it was a sentence no cart prints.

The break is `\n`, which is what `src/import/RomExtractorGen2.lua:2558` decodes
the cart's own `line` ($4e) into, so these read exactly as an extracted line
would. Each message now goes through `Strings`, which is what the rest of this
file already does (`Coins scattered\neverywhere!` at `:1758`, `But nothing\nhappened.`
at `:1946`), so a translation mod can reach them.

### SpikesText is deliberately not in this change

`SpikesText` (`data/text/battle.asm:974`) is the sixth line I found, and it is
the interesting one:

```
	text "SPIKES scattered"
	line "all around"
	cont "<TARGET>!"
```

Three rows, the third a `cont`. **The battle message path cannot tell it.**
Engine events set `self.message` straight from `event.text`
(`src/ui/gen2/BattleState.lua`), and `printMessage` draws
`Chrome.wrap(self.message, TEXT_WIDTH)` capped at `TEXT_ROWS`, so a third row
is cut, not scrolled. `paginate`/`showPages` exists and handles `\v` correctly,
but only `win-text` and a couple of prompts go through it.

I had written the Spikes fix, and the box check below is what caught it: the
line wraps to 3 rows, so the target's name would have silently vanished from
the screen. So Spikes keeps its current wording, with the reason recorded
beside it. Wiring engine messages through `showPages` looks like the real fix,
and I did not want to widen a text change into that. Happy to take it as a
follow-up if you want it.

### Tests

`tests/engine/gen2_battle_text_matches_cart.lua` drives a real turn per case
and compares the emitted line against the label. `tests/run_engine.lua` picks
it up on its own: **288/288 suites, ALL TESTS PASSED**.

Against the text as it stood the new file is **5/11, 6 failures**, so it pins
the change rather than describing it. Its last case wraps every changed line
through the real `Chrome.wrap` at the box's own width and holds it to the two
rows `printMessage` draws.

`tests/gen2_battle_ui_test.lua:178` asserted the old not-very string and moves
with it (288/288).

whole of `tests/` and `tests/engine/` against `dev` at `087a2751` in a second
worktree: identical five, so nothing here regressed them.

### The bigger thing behind this

Hand-written lines drift because Gen 2 has no equivalent of
`src/core/RomText.lua`. There is no `text` section in `rom_manifest_gold.json`
or `_silver.json` (Red/Blue carry 2595 labels, Yellow 2710), no `extractText`
in `RomExtractorGen2`, and `data.text` is never assigned on the Gen 2 path.
Its extracted text is keyed by `bank:address` into `data.gen2Text`, and only
the overworld VM reads it. So all 132 `kind = "message"` emits in the two Gen 2
battle files are engine-authored, and the ones I did not touch have their own
smaller drifts (`PerishCountText` is right next door in the asm and prints as
one row here).

Happy to open that as its own issue with the numbers, and to take a run at the
manifest walk plus `extractText` if you like the shape. It would also close out
the four concatenations in #1348, which is marked completed but still reads as
reported on `dev`.

### DIFF
--- a/src/battle/gen2/Battle.lua
+++ b/src/battle/gen2/Battle.lua
@@ -1234,10 +1234,15 @@ function Battle:dealDamage(attacker, defender, damage, opts)
   if opts.critical then
     self:emit({ kind = "message", text = "A critical hit!" })
   end
+  -- SuperEffectiveText / NotVeryEffectiveText (data/text/battle.asm:603,608).
+  -- The cart breaks both across the box's two lines and hyphenates "super-"
+  -- to do it, and the not-very line ends on the single ellipsis glyph Gold's
+  -- charmap carries at $75, not three periods.
   if opts.effectiveness and opts.effectiveness > 10 then
-    self:emit({ kind = "message", text = "It's super effective!" })
+    self:emit({ kind = "message", text = Strings("It's super-\neffective!") })
   elseif opts.effectiveness and opts.effectiveness < 10 then
-    self:emit({ kind = "message", text = "It's not very effective..." })
+    self:emit({ kind = "message",
+      text = Strings("It's not very\neffective…") })
   end
   if endured then
     self:emit({ kind = "message",
@@ -1391,7 +1396,9 @@ function Battle:useMove(attacker, defender, moveId)
 
   if not (charging or rampaging or rolling) then
     if move and (move.pp or 0) <= 0 then
-      self:emit({ kind = "message", text = "No PP left for this move!" })
+      -- BattleText_TheresNoPPLeftForThisMove (data/text/battle.asm:315).
+      self:emit({ kind = "message",
+        text = Strings("There's no PP left\nfor this move!") })
       return
     end
     if move then move.pp = (move.pp or 1) - 1 end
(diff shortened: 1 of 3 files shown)


bryanthaboi:
I like what you're saying. Go for it

Issue #1665, opened 2026-08-21
gold has no bide lock-in, so a mid-bide switch of move strands the counter


bryanthaboi:
`Battle:lockedInMove` (src/battle/gen2/Battle.lua:3774) and `Battle:forcedMove` (:3785) cover rollout, rampage and encore but never bide, so the player menu and the ai can both pick something else while bide is storing. `bideTurns` then stays nonzero for the rest of the battle and `bideStored` keeps banking damage, so the next bide unleashes an inflated hit. the cart locks both sides (engine/battle/core.asm:574-576 and :5648-5651) and gen 1 already does it at src/battle/BattleState.lua:2003. found during the aug 21 research batch, details in .bazinga/august21p2/fix_reports/1481-1485-adjudicate.md

colsonrice:
Taking this one alongside #1664.

Confirmed on `dev` at `087a2751`: `Battle:lockedInMove` returns only
`state.rolloutLock` and `state.rampageMove`, so nothing holds the menu on bide,
and `bideTurns` never clears once the player picks something else.

`BattleState:fightLockedAction` looks like the Gen 1 template you meant:
`if battler.bideTurns then return { special = "bide" } end`. I will mirror that
rather than invent a second shape.

Planning one PR covering both issues, with a headless test per issue.

PR #1674, opened 2026-08-22
Give bide its two carve-outs on Gold

### LINKED ISSUE #1664: gold bide spends pp every turn instead of only the turn it is selected
`src/battle/gen2/Battle.lua:1392` exempts charging, rampage and rollout from the pp decrement but not bide, so a bide that runs three turns costs 3 pp instead of 1. the cart's doturn carve-out mask includes `1 << SUBSTATUS_BIDE` (engine/battle/effect_commands.asm:977-979), and StoreEnergy jumps past doturn on every turn after the first (engine/battle/move_effects/bide.asm:65-68). found during the aug 21 research batch, details in .bazinga/august21p2/fix_reports/1481-1485-adjudicate.md

colsonrice:
Closes #1664 and #1665. They land together because the lock needs a record of
which move it is holding, and the PP exemption reads the same field.

### #1665, the lock

`Battle:lockedInMove` returned only `state.rolloutLock` and `state.rampageMove`,
so nothing held either side on bide. The menu and the ai could both pick
something else while it stored, `bideTurns` never reached zero, and
`bideStored` kept banking for the rest of the battle, so the next bide
unleashed an inflated hit.

The cart skips `MoveSelectionScreen` while `SUBSTATUS_BIDE` is set:

```
	ld a, [wPlayerSubStatus3]
	and 1 << SUBSTATUS_BIDE
	jr nz, .locked_in          ; engine/battle/core.asm:574-576
```

and `CheckEnemyLockedIn` masks the same bit alongside charged and rampage
(`:5648-5651`). Gen 1 already holds the menu, at
`BattleState:fightLockedAction`, so this is the Gen 2 side catching up rather
than a new idea.

`state.bideMove` is the record. It is set beside `bideTurns` and cleared with
it, which is the shape `rampageMove` already uses.

### #1664, the PP

The carve-out in `useMove` read `if not (charging or rampaging or rolling)`,
with no bide term, so a bide that stored for three turns cost 3 PP.

`doturn` returns before `.consume_pp` when the user carries the bit:

```
	ld a, [de]
	and 1 << SUBSTATUS_IN_LOOP | 1 << SUBSTATUS_RAMPAGE | 1 << SUBSTATUS_BIDE
	ret nz                     ; engine/battle/effect_commands.asm:978-979
```

and StoreEnergy's continuation skips straight to `unleashenergy_command` on
every turn after the first (`engine/battle/move_effects/bide.asm:61-62`). So
the whole run costs the one PP the opening turn spent, the same way a rampage
costs one.

The bit is set by StoreEnergy, which runs after `doturn`, so the opening turn
still pays. That ordering is why the guard tests `bideTurns` rather than the
move id alone.

### Tests

`tests/engine/gen2_bide_lock_and_pp.lua` drives real turns and asserts the run
costs one PP, that the menu is held while bide stores, that a menu naming
another move cannot strand the counter, and that the lock lifts with the
unleash. `tests/run_engine.lua` picks it up on its own: **289/289 suites, ALL
TESTS PASSED**.

Against the code as it stood it is **4/11, 7 failures**, including the
stranding itself:

```
FAIL bide spent one PP across 4 turns, not one per turn (got 6, want 9)
FAIL lockedInMove holds the user on bide while it stores (got nil, want BIDE)
FAIL the counter is cleared rather than stranded (got 3, want nil)
```

`tests/gen2_battle_test.lua` stays at 705 checks, 0 failures.

`scripts/lint.sh --gate` is clean, 0 warnings and 0 errors across 482 files.

whole of `tests/` and `tests/engine/` against `dev` at `15fc04e0` in a second
worktree: identical five, so nothing here regressed them.

### One thing I did not touch

Bide's turn count. With `Effects.bideTurns` rolling 2 or 3 and the decrement
happening on the turns after the opening one, a run takes 4 turns end to end
in this port. Whether that matches `wPlayerCharging` on the cart is a separate
question from either issue here, so I left it alone rather than fold a third
change in.

### DIFF
--- a/src/battle/gen2/Battle.lua
+++ b/src/battle/gen2/Battle.lua
@@ -1389,7 +1389,15 @@ function Battle:useMove(attacker, defender, moveId)
   -- free of PP and obedience in exactly the same way.
   local rolling = state.rolloutLock == moveId
 
-  if not (charging or rampaging or rolling) then
+  -- The doturn command returns early when the user's substatus carries
+  -- SUBSTATUS_BIDE (effect_commands.asm:978-979), and StoreEnergy's
+  -- continuation skips straight to unleashenergy on every turn after the
+  -- first (move_effects/bide.asm:61-62).  So the whole run costs the one PP
+  -- the opening turn spent, the same way a rampage costs one.
+  local biding = def.effect == "EFFECT_BIDE"
+    and state.bideMove == moveId and (state.bideTurns or 0) > 0
+
+  if not (charging or rampaging or rolling or biding) then
     if move and (move.pp or 0) <= 0 then
       self:emit({ kind = "message", text = "No PP left for this move!" })
       return
@@ -2133,6 +2141,8 @@ Battle.MOVE_EFFECTS.EFFECT_BIDE = function(self, attacker, defender, def, moveId
   if not state.bideTurns then
     state.bideTurns = Effects.bideTurns(self.random)
     state.bideStored = 0
+    -- what the lock below holds the user on, the way rampageMove does
+    state.bideMove = moveId
     self:emit({ kind = "message",
       text = self:monName(attacker) .. " is storing energy!" })
     return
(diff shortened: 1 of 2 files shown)


colsonrice:
Closing this. `c23f82ef` covers both issues, and covers more than this did:
`cancelBide` on a switch or an item, the separate `fightLockedMove` for the
ordering against encore, and the enemy twin are all things this PR did not do.

#1673 is still open and unaffected. The five battle lines it changes are
untouched on `dev`, and it still applies cleanly at `51bc4c74`.

PR #1680, opened 2026-08-22
Give Gen 2 the ROM text table Gen 1 has had all along

colsonrice:
Following up on #1673, where you said go for it.

Gold and Silver had no label-keyed string table at all. The manifests carried
no `text` section and `RomExtractorGen2` had no `extractText`, so **every**
call through `src/core/RomText.lua` fell back to the literal written beside
it. The only Gen 2 text in a cache was the script text in
`data/generated/text.lua`, keyed by `bank:address` for the overworld VM,
which nothing can look a battle line up in.

`Data:load` does list `text` among its modules, but only `Game:load` calls it
and `Game2:load` never assigned the field, so on a Gold session `data.text`
was either absent or whatever a Gen 1 session earlier in the same process had
left there. Either way it could not answer a Gen 2 label.

|  | Red/Blue | Yellow | Gold | Silver |
|---|---|---|---|---|
| `text` section, before | 2595 labels | 2710 | none | none |
| `text` section, after | unchanged | unchanged | **889** | **889** |

### The three pieces

**1. The manifest names the labels.** `make_gold_manifest.py` walks
`data/text/`'s five dialogue files the way `make_rom_manifest.text_metadata`
walks pokered's, and embeds each label's symbol. The other files there are
character tables rather than strings: `dakutens.asm`, `name_input_chars.asm`
and `mail_input_chars.asm` are keyboard layouts and
`unused_gen1_trainer_names.asm` is a dead Gen 1 leftover. Decoded as text they
come out as keyboard rows and kana runs, so `TEXT_SOURCES` leaves those files
out by name.

`BattleText::` is excluded for a different reason: it is a bank anchor whose
own comment says `used only for BANK(BattleText)`, and it shares an address
with the first real label under it, so taking it would decode that
neighbour's string twice. The walk requires a label to be followed by
something other than another label.

`make_silver_manifest.py` needed no change. It inherits the list and
re-resolves addresses from `pokesilver.sym`, and all 889 are present there
too.

**2. `RomExtractorGen2:extractText`** decodes them into
`data/generated/rom_text.lua`. The mechanism is the one `extractOakSpeech`
already used for `_OakText1-7`: resolve the label, decode from the cart, key
by name. What is new is that the list comes from the manifest rather than
being written out in Lua, so all of `data/text/` arrives instead of seven
strings. It is `rom_text` rather than `text` because `text.lua` is already the
address-keyed script table.

**3. `Game2:load` assigns it to `self.data.text`**, which is what makes the
existing shared `RomText` helper work on Gold and Silver at all.

### The name slots it also had to fix

`decodeGen2Text` dropped `<USER>`, `<TARGET>` and `<ENEMY>`, because the
charmap gives them angle-bracket names and the decoder skips anything shaped
like a control glyph. They are not control glyphs: `PlaceMoveUsersName`,
`PlaceMoveTargetsName` and `PlaceEnemysName` (`home/text.asm:302`, `:307`,
`:327`) substitute a battler's name as the line prints. Skipped,
`SubTookDamageText` decoded as `"The SUBSTITUTE / took damage for"` with
nothing after it. They now decode to `{USER}` / `{TARGET}` / `{ENEMY}`, the
shape `RomText` already fills in argument order, the same as Gen 1.

That was the whole gap. Across all 889 labels the decoder produces **zero**
`{BYTE:..}` escapes, so nothing else in `data/text/` is unmapped.

### Verified against a real cartridge

I built pokegold from pret at `656583c` with rgbds 1.0.3. Both editions come
out byte-exact:

```
pokegold.gbc    d8b8a3600a465308c9953dfa04f0081c05bdcb94
pokesilver.gbc  49b163f7e57702bc939d642a18f591de55d92dae
```

which match `GameVersion` and both manifests' `romSha1`, and `make compare`
passes. Running the real `extractText` against that ROM produces 889 strings,
including:

```
SuperEffectiveText   "It's super-\neffective!"
StartPerishText      "Both POKéMON will\nfaint in 3 turns!"
PlayerHitTimesText   "Hit {NUM} times!"
SubTookDamageText    "The SUBSTITUTE\ntook damage for\v{TARGET}!"
WantsToBattleText    "{ENEMY}\nwants to battle!"
```

The first three are character-for-character the lines I hand-wrote in #1673,
now coming off the cartridge instead.

Before regenerating anything I checked that `make_gold_manifest.py` reproduces
the committed manifest byte for byte from that checkout, so the diff is only
this change: one key added, 790 symbols added, **nothing existing altered**.

### Tests

`tests/gen2_rom_text_test.lua`, registered in `tests/run_tests.lua`:

- both manifests carry a `text` section, and every label in it resolves to a
  symbol. A label the list names but the symbols cannot place would fail an
  import at the Dialogue stage rather than at generation time, so the pairing
  is asserted here.
- the two editions name the same labels in the same order
- the excluded keyboard tables and the `BattleText` anchor stay out
- `RomText` fills `{USER}` / `{TARGET}` / `{ENEMY}`, and an absent label still
  falls back
- a cache-fed section compares real decoded strings to pokegold's
  `data/text/battle.asm`, skipping cleanly when there is no cache, in the
  shape `gen2_battle_ui_test` already uses for `GOLD_CACHE`

Against the manifests as they stood it is 19/23. `run_engine` is 290/290, ALL
TESTS PASSED. `scripts/lint.sh --gate` is clean, 0 and 0 across 482 files.

`gen2_diploma_test` and `gen2_trade_gfx_test` both assert the literal source
text `local STAGE_COUNT = 26`, which is exactly the guard that should fire
when someone adds a stage. Both move to 27.

whole of `tests/` and `tests/engine/` against `dev` in a second worktree:
identical five.

### What I checked before sending this

- **The decoder change touches nothing else.** `decodeGen2Text` is shared with
  the script text, so I ran `extractScriptsAndText` against the built ROM with
  and without the `NAME_SLOT` lines and diffed the two `text.lua` files: 3046
  entries, **zero** differing lines. Those three bytes never appear in map
  script text, which is what you would expect from battle-only substitutions.
- **No label collides across generations.** `data.text` is the shared `Data`
  table, and Gen 1's `BattleState` and `PartyMenu` read 12 labels off it. None
  of the 12 appears in the Gen 2 list, and Gold has its own
  `src/ui/gen2/PartyMenu.lua`, so nothing reads the wrong table.
- **Gen 1 cannot inherit the Gen 2 table.** `Game:load` calls `Data:load()`
  unconditionally, so a Red session after a Gold one repopulates `text` from
  its own cache.
- **Both manifests still reproduce byte for byte** from the same pokegold
  checkout after the change, and the branch merges clean on `dev` at
  `478e3bf8` with the engine tier at 291/291 on the merged result.

### Three things I left alone

1. **No message is routed through `romText` yet.** This is the enabling half.
   Routing the Gen 2 battle emits is mechanical and reviews better on its own,
   the way #1559 landed for Gen 1. Happy to send it next.

2. **The new table is not moddable.** `Schemas.lua` maps a Gen 2 mod's `text`
   key to `gen2Text`, so `data.text` has no registry entry yet. That looked
   like your call rather than mine, since it decides whether a mod overrides
   engine text by label.

3. **Silver has no `VERSION_REQUIRED_FILES_OVERRIDE` entry**, so
   `requiredFilesFor` falls through to the Gen 1 `REQUIRED_FILES` list and a
   Silver cache is checked for things a Gen 2 extract does not write. That
   looks pre-existing and unrelated, and fixing it inside a text change felt
   wrong, so the new marker went in the `gold` list only. Flagging it because
   it means the marker will not re-import a Silver cache.

### DIFF
--- a/src/core/Game2.lua
+++ b/src/core/Game2.lua
@@ -943,6 +943,12 @@ function Game2:load()
   self.data.gen2Scripts = loadGenerated("data/generated/scripts.lua")
   self.data.gen2StdScripts = loadGenerated("data/generated/std_scripts.lua")
   self.data.gen2Text = loadGenerated("data/generated/text.lua")
+  -- The engine's own strings, keyed by the disassembly's label.  gen2Text
+  -- above is the script text and is keyed by bank:address for the overworld
+  -- VM, so the two are different tables and both are loaded.  This one is
+  -- what src/core/RomText.lua reads, which is why it lands on `text`: that
+  -- helper is shared with Gen 1 and looks up data.text[label].
+  self.data.text = loadGenerated("data/generated/rom_text.lua") or {}
   -- data/generated/events.lua: the side tables a script command NAMES rather
   -- than carries -- the phone book, the in-game trades, the elevator's floor
   -- labels, the decoration descriptions.  Keyed for World's own `eventTables`
--- a/src/import/RomExtractorGen2.lua
+++ b/src/import/RomExtractorGen2.lua
@@ -27,7 +27,7 @@ RomExtractorGen2.__index = RomExtractorGen2
 -- scripts, pokemon, moves, items, marts, encounters, trainers, pokedex,
 -- landmarks, intro movie, menu gfx, title, credits, diploma, trade animation,
 -- audio, stubs
-local STAGE_COUNT = 26
+local STAGE_COUNT = 27
 local Opcodes = require("src.script.gen2.Opcodes")
 
 -- BG palette slots inside one loaded 8-palette set (constants/tileset_constants.asm
@@ -134,6 +134,18 @@ local TEXT_NO_GLYPH = {
   [0x05] = true, [0x07] = true, [0x0a] = true, [0x0b] = true, [0x0d] = true,
   [0x0e] = true, [0x0f] = true, [0x10] = true, [0x11] = true, [0x12] = true,
   [0x13] = true, [0x15] = true,
+}
+
+-- The three runtime name slots.  PlaceMoveUsersName, PlaceMoveTargetsName and
+-- PlaceEnemysName (home/text.asm:302, :307, :327) swap these for a battler's
+-- own name as the line prints, so they are markers rather than glyphs.  They
+-- decode to the same shape Gen 1 uses, which src/core/RomText.lua already
+-- fills in argument order.  Dropped, SubTookDamageText read "took damage
+-- for" with nothing after it.
+local NAME_SLOT = {
+  ["<USER>"] = "{USER}",
+  ["<TARGET>"] = "{TARGET}",
+  ["<ENEMY>"] = "{ENEMY}",
 }
 
 local ROOF_TILES = 9
@@ -2596,6 +2608,8 @@ function RomExtractorGen2:decodeGen2Text(bank, address, charmap, buffers)
         out[#out + 1] = ch
       elseif ch == "<……>" or b == 0x56 then
         out[#out + 1] = "……"
+      elseif NAME_SLOT[ch] then
+        out[#out + 1] = NAME_SLOT[ch]
       elseif not ch then
         out[#out + 1] = ("{BYTE:%02X}"):format(b)
       end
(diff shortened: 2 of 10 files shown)


bryanthaboi:
This is good yeah. So it's there but won't affect anything at all yet is what you're saying?

colsonrice:
yeah nothing reads it yet, it's just the table sitting there.

only visible thing is gold/silver caches re-import once off the new marker, and there's one extra stage on the import progress bar.

routing pass is the next bit, same as #1559 for gen1. swap the hardcoded lines in Battle.lua and gen2/BattleState.lua over to romText with their labels.

PR #1746, opened 2026-08-23
engine: generic seams for mod-driven multiplayer modes

campavao:
Five small engine additions that let a mod drive its own overworld actors and run a link battle over a connection it already has. I hit these building an overworld multiplayer mode, but none of them mention it — each is useful to anything with a self-scheduled actor, an already-paired transport, or its own new-game flow.

This is a Route B change (it adds a hook and an event), so it carries an RFC: **`docs/rfcs/0014-mod-driven-actors-and-adopted-link-sessions.md`**.

### What it adds

WorldAPI handles gain `stepNow`, `canStep`, `placeAt`, `isMoving` and `setPassable`. The existing `scriptMove` queues onto `OverworldState.scriptMoves`, and a non-empty `scriptMoves` is how the overworld knows a cutscene is running, so it gates `handleInput` — anything animated that way freezes the player's controls for as long as it walks. That's right for Oak marching into his lab and wrong for an actor moving on its own schedule. `stepNow` sets the same per-tile state without the queue. It skips the collision check on purpose: the caller is replaying a move that was already decided somewhere else, and re-judging it here would let the two copies disagree about where the actor is. `canStep` is the separate opinion for callers that want it.

`world.talk` is the A press on an object, raised before the map's text tables get it. An object from `spawnNpc` has no `TEXT_*` id, so the vanilla path has nothing to say for one; a mod that owns the object wraps this and doesn't call `next`. Everything else falls through unchanged, and an object mid-step raises no hook, same as vanilla.

`LinkState.newFromSession` adopts a transport that's already paired and skips the connect UI, keeping the hello and fingerprint exchange intact. That lets a mode tunnel a battle through a connection it already has instead of opening a second one. When the battle ends, `link.battle_ended` reports `{ result, myParty, theirParty, peerName, role }`. The party copies are the useful part — cable rules leave the real party untouched, so by the time the state unwinds there's nowhere else to read what the fight cost.

`Game:startNewGame(opts)` is the title's NEW GAME closure made callable, with `opts.intro = false` to land straight in the world. `CodeEntry.new` takes an optional `{ length, charset }` so the slot-scrub widget can carry an address or a room code as well as a link code.

### Backward compatibility

Every item is a new name or a new optional argument. No existing name, payload, signature or default changes, so existing mods do nothing. The v1 surface is untouched — `content.X:register/override/get`, `events:on`, `hooks:wrap`, `mod.log`, `mod:read`, the manifest v1 fields and `pokemon.before_give` all behave as before, and `mods/example_mew_starter` (api 1, `category = "GAMEPLAY"`, whole-species copy) still loads unchanged, which `run_modkit` proves on every run.

### Parity

Both seams are guarded. `link.battle_ended` is behind `Runtime.wants`, so an unsubscribed build allocates no payload — I'd missed that on the first pass and fixed it in the third commit, since Route B is explicit about it and every other emit in the engine already guards. `world.talk` passes a file-local fallthrough rather than a closure built per A press, so an unhooked press allocates nothing new either.

Both unit tests drive the real engine path rather than the hook in isolation, and both assert the unhooked build first, so they fail if the seam is deleted *and* if it changes vanilla behaviour. The structural side needed nothing — `gate_hooks` and `gate_events` walk the live catalog, so both names were parity-gated as soon as their call sites existed.

No registry or schema field is added, so `gen_registry_docs` has nothing to emit here. Running it does surface pre-existing drift in `docs/modding/reference/registries.md` (`timeFishGroups`, an `objects` refinement) from earlier `Schemas.lua` edits that were never regenerated — unrelated to this branch, so I left it alone rather than bundle it in.

### Tests

`gate_meta_coverage` 208/208, `world_talk` 9/9, `link_battle_ended` 16/16, `run_modkit` 26/26, `run_link_tests` green. `run_engine` is 289/292 — the three failures (`audio_device_reset`, `quit_thread_shutdown`, `title_zone_seams`) fail the same way on an unmodified `origin/dev` worktree, so they aren't from this branch.

One heads-up: #1734 also touches `src/world/WorldAPI.lua` and `docs/modding.md`. Different regions, but whichever lands second will want a look.

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -265,6 +265,80 @@ Red exposes FLY separately because it requires a destination picker:
 and `mod.world:flyTo(mapId)` accepts only a visited destination from the native
 Fly town list. Gold does not expose these two methods yet.
 
+## Self-driven world actors
+
+`mod.world:spawnNpc()` returns a handle whose `scriptMove` queues onto the
+overworld's scripted-movement list. A non-empty list is how the overworld
+knows a cutscene is running, so it gates player input for as long as the
+actor walks -- right for Oak marching to his lab, wrong for an actor that
+moves on its own schedule (a networked player's ghost, an ambient walker).
+Five handle methods drive one without that lockout:
+
+```lua
+local ghost = mod.world:spawnNpc({ map = "ROUTE_1", x = 5, y = 7,
+                                   sprite = "SPRITE_RED" })
+if ghost:canStep("up") then ghost:stepNow("up") end   -- one tile, now
+if not ghost:isMoving() then ghost:placeAt(9, 3, "down") end  -- snap, no walk
+ghost:setPassable(true)                               -- walk-through
+```
+
+`stepNow(dir)` sets the same per-tile state `scriptMove` does, minus the
+queue. It deliberately does **not** check collision: a caller replaying a
+move that was already decided elsewhere (validated on a peer's machine, or
+authored) would let the two copies disagree about where the actor is if this
+re-judged it. Ask `canStep(dir)` first when you do want the map's opinion.
+`placeAt(x, y, facing)` snaps with no animation and clears any step in
+flight, for a warp arrival or a resync too far gone to walk off.
+`isMoving()` lets a driver pace itself instead of stomping a move already
+running. `setPassable(flag)` is the flag `Collision.occupied` skips (the
+engine's own user is Yellow's companion Pikachu); a passable object still
+draws and can still be talked to.
+
+An object spawned this way carries no `TEXT_*` id, so the vanilla talk path
+has nothing to say for it. The **`world.talk`** hook is the A press on an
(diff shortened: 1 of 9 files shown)


campavao:
hey @bryanthaboi! first off, huge fan of this project in general! Thanks for making it, and making it so easy to contribute!

the TLDR behind these changes is that I'm trying to make a Pokemon Battle Royale mod (see [here](https://github.com/campavao/kanto-battle-royale)) and needed some additional networking stuff in place in order to make that work. I think I am actually able to get the mod to work without these changes to the core, but it's a bit flimsy. I'm gonna make the mod in a way that can support w/o this change, and w/ it tho

this might be helpful for others who are trying to build more multiplayer featured mods, so figured I'd keep this open and work with ya to get it merged (if you're open to it)

lmk what you think and if there's anything you need/want from my end!

PR #1764, opened 2026-08-24
mod API: report caught Pokémon storage success

BoLayerDev:
## Summary

- add a required boolean `stored` field to `pokemon.caught`
- report `true` only after the exact caught Pokémon enters the party or a box
- report `false` when a Gen 1 catch succeeds but both party and box storage reject it
- preserve the legacy `destination` route field and all existing catch results
- document the shared Gen 1/Gen 2 contract

## Compatibility

This is an additive payload field. Existing listeners and vanilla behavior remain unchanged. Consumers that require confirmed storage can use `ev.stored == true`, which fails closed on older engines where the field is absent.

Gen 2 already emits only after successful insertion, so it reports `stored = true`.

## Tests

ROM-free focused/adversarial suite:

- party insertion
- successful box insertion
- full-box rejection
- object identity
- event ordering
- absent and throwing listeners
- ordinary and Safari shared paths
- save-state behavior
- Gen 2 parity

Local results:

- `pokemon_caught_stored_contract.lua`: 141/141
- `gate_hooks.lua`: 493/493
- `gate_events.lua`: 529/529
- `gate_gen2_mod_api.lua`: 943/943
- `gate_meta_coverage.lua`: 206/206
- broad engine runner: 288/293; the exact base is 287/292 with the same five Windows/environment-sensitive failures
- `git diff --check`: pass

The change contains no ROM, save, cache, or extracted game data.

### DIFF
--- a/docs/mod-api-gen2-compat.md
+++ b/docs/mod-api-gen2-compat.md
@@ -563,10 +563,13 @@ gains a field instead of the name gaining a prefix.
   happen either way.
 - *The catch and the evolution:* `pokemon.caught`, `pokemon.evolved`; hook
   `evolution.check`. `src/ui/gen2/BattleState.lua:pushCaught` emits
-  `pokemon.caught` once the mon is in the party or the box, and
-  `src/core/gen2/Evolution.lua` emits `pokemon.evolved` from `apply` and wraps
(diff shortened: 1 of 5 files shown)


castdrian:
this needs a rebase.

PR #1798, opened 2026-08-25
engine: battle.style and catch.nickname rule hooks

campavao:
Two decisions a Red/Blue/Yellow battle takes from the OPTION screen and the cart, made hookable so a game mode can take them instead. I hit both building a battle-royale mode, but neither mentions it — a tournament ruleset, a Nuzlocke, a randomizer, a speedrun practice mod all want the same two switches.

This is a Route B change (it adds two hooks), so it carries an RFC: **`docs/rfcs/0015-battle-rule-hooks.md`**.

### What it adds

`battle.style` wraps the SHIFT/SET read at the moment the foe's Pokémon faints and the engine would offer the free switch. Today that read is inline (`save.options.battleStyle`), so the only way a mode can force SET is to write the player's saved preference — and that leaks into their real playthrough the first time anything calls `Game:writeOptions`, which the speed hotkey does on every press. The hook returns `"set"` or `"shift"` for this battle; anything else reads as the vanilla answer, so a wrapper that returns nothing by mistake cannot change the rule. The player's preference is never written.

`catch.nickname` wraps the `AskName` prompt after a capture. The script-gift path already has exactly this seam — `pokemon.before_give`'s `gift.nickname` — so the two ways a Pokémon joins the party now answer the same question the same way: `false` keeps the species name with no prompt, a string is the nickname (clipped to the naming grid's ten characters; an empty string names nothing, like the grid's own empty entry), anything else asks as today.

Both live as methods on `BattleState` (`battleStyle`, `offerNickname`) called from the paths that previously inlined the logic — methods on purpose, so a mod or a compatibility shim can tell a seam engine from a stock one by name.

### Why a mod can't do this from outside

Both reads are mid-function, with no name to wrap. A mod's alternatives today are writing the player's options (leaks, above) or driving the yes/no box after it opens (the prompt has already flashed, and the timing is fragile). Neither survives contact with a real player's save.

### Backward compatibility

Additive: two new method names, no existing name, payload, signature or default changes. The v1 surface is untouched — `mods/example_mew_starter` (api 1) still loads unchanged, which `run_modkit` proves on every run.

### Parity

Both call sites are guarded by `Runtime.wantsHook` and both vanilla links are file-locals, so an unhooked build runs the branch exactly as before and allocates nothing it did not allocate before. `tests/modkit/cases/battle_rule_hooks.lua` drives both hooks through the public mod API and asserts the no-mod answers first — the OPTION row, and the cart's AskName — so it fails if the seam is deleted *and* if it changes vanilla behaviour. The catalog scans call sites, so `gate_hooks` parity-gated both names as of this branch.

### Tests

`gate_hooks` 521/521, `gate_meta_coverage` 211/211, `run_modkit` 28/28 suites.

One heads-up: #1764 also touches `src/battle/BattleState.lua` and `docs/modding.md`, and #1799 (RFC 0018, `catch.party_full`) edits the same capture path — whichever lands last is a trivial rebase.

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -505,6 +505,42 @@ Menu choices and moves use the same engine methods as the native controls;
 their mutable logic. Tutorial, link, forced, stale, and covered battle states
 refuse core intents. Use `mod.input` for ordinary text advance.
 
+## Battle rule hooks
+
+Two decisions the OPTION screen and the cart make for the player, which a game
+mode can make instead (RFC 0015). Neither writes the player's saved
+preference, so a mode can hold a rule for as long as it is active and hand the
+player's own setting back untouched.
+
+`battle.style` wraps the SHIFT/SET read at the moment the foe's Pokémon faints
+and the engine would offer a free switch:
+
+```lua
+mod.hooks:wrap("battle.style", function(next, battle)
+  if myMode.active then return "set" end   -- no "will you change POKéMON?"
+  return next(battle)                      -- the OPTION row, as today
+end)
+```
+
+Return `"set"` or `"shift"`; anything else reads as the vanilla answer.
+
+`catch.nickname` wraps the `AskName` prompt after a capture (party or box),
+the same question `pokemon.before_give`'s `gift.nickname` already answers for
+script gifts:
+
+```lua
+mod.hooks:wrap("catch.nickname", function(next, mon, ctx)
+  -- ctx = { battle = <BattleState>, name = <display name>, game = <Game> }
+  if myMode.active then return false end        -- keep the species name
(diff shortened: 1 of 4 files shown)


castdrian:
please try to avoid backwards merging from origin and rebase your branch instead to avoid polluting the commit history

campavao:
> please try to avoid backwards merging from origin and rebase your branch instead to avoid polluting the commit history

Ah sorry! Can rebase and squash my commits rq

campavao:
Done — rebased onto `dev` and force-pushed, so the backwards-merge commit is gone and it's back to the single RFC commit. Thanks for the review!

PR #1808, opened 2026-08-25
fix(mobile): in-process launcher return for Android and iOS

1Jamie:
Stop joining ChipAudio on every EXIT GAME, always rebuild the Gen1 Game module on Play-again, and treat iOS like Android for return-to-launcher so release/mobile LOVE restarts no longer wipe Game.load or crash on quit.

### DIFF
--- a/main.lua
+++ b/main.lua
@@ -384,12 +384,14 @@ function bootGame(version, cartId)
     Game = require("src.core.Game2").new()
     Game:load()
   else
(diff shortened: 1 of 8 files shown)


1Jamie:
please for the love of god, if you test anything, test this one on ios please before merging or read through it since you know ios better than i do @bryanthaboi  lol
I went through the docs apple provides and what lov recommends.

PR #1835, opened 2026-08-26
engine: pokemon.level_visible, a level a mode can take off the screen (RFC 0019)

campavao:
Adds `pokemon.level_visible` — a guarded seam that lets a mod take a Pokémon's LEVEL off the screens that print it, without moving anything else on them.

RFC: `docs/rfcs/0019-pokemon-level-visible.md`

## Why

A level is printed on four Gen 1 surfaces — both battle healthboxes, the party rows, and both status pages — and every one prints it unconditionally. With no seam, a mode that wants the number gone has two options, and both are bad:

- paint over the engine's own pixels from `render.hud` — four rectangles, a background shade to match, and the palette flashes and healthbox slide to survive; or
- monkey-patch the render modules from inside the sandbox. This *works* — `require` hands a mod the engine's own table — and is exactly what `CONTRIBUTING-mods.md` tells mods not to do.

The motivating case is a battle royale that scales every party to a shared rung that rises with its fog. The number is the same for everyone and changes on a clock, so it is never news — and it reads as a threat it is not: a Lv37 opponent looks dangerous to a player who has not worked out that their own team is Lv37 too. A randomizer keeping an encounter unreadable, a challenge run that forbids level-checking, and a blind Nuzlocke all want the same switch, and none of them should have to learn where the `<LV>` glyph lives.

## Shape

It follows the presentation predicates already on the battle screen — `battle.status_hud_visible`, `battle.bottom_ui_visible`, `battle.caught_marker_visible`: consulted behind `Runtime.wantsHook`, defaulting to visible, only an explicit `false` suppresses.

```lua
mod.hooks:wrap("pokemon.level_visible", function(next, mon, ctx)
  -- ctx = { where = "battle.enemy" | "battle.player" | "party" | "summary",
  --         game = <Game> }
  if myMode.active and ctx.where ~= "party" then return false end
  return next(mon, ctx)
end)
```

Two deliberate departures from those siblings:

- **Not named `battle.*`**, because a level is not a battle-only readout.
- **It carries the surface that asked.** The number means different things on different screens: on a battle HUD an opponent's level is information about *them*, on the party and status screens your own level is information about *you*. A mode can hide one and keep the other, and the motivating mode does.

`src/ui/LevelDisplay.lua` holds the single definition of "visible", so the four call sites are a one-line guard each instead of four copies of the same five lines that can drift apart.

## What does not change

No layout moves. Each site keeps its own hand-rolled PrintLevel rule (`home/pokemon.asm:335-345` — the `<LV>` tile, then the digits, with a level of 100 writing its third digit back over the tile); it just asks first.

Two details worth calling out in review:

- **A status condition still replaces the level on a healthbox**, exactly as in the cart, so hiding a level never hides `PSN` or `BRN`. The guard is an `elseif` on the existing status branch rather than a wrapper around it.
- **On status page 2 the `<to>` arrow is hidden with the level it points at** — an arrow with nothing after it is half a sentence. The EXP-to-next-level figure beside it is a separate field and still prints.

`wantsHook` is checked before `call`, which matters because two of the four sites are on a per-frame draw path: a build with no mod wrapping the hook never reaches the call and its pixels are unchanged.

## Scope

**Gen 1 only.** The Gen 2 screens (`src/ui/gen2/{PartyMenu,SummaryMenu,BoxMenu,HallOfFame}.lua`) and the Gen 1 PC box list keep their own readouts and do not consult the hook — in the box list the level is part of a row label string rather than a drawn field, which is a different shape of change.

Those are mechanical follow-ups, held back so this stays reviewable against screens that can actually be exercised here. The limitation is stated in `docs/modding.md` beside the hook and in the RFC, so a mod author reads it before depending on it rather than after. Happy to fold either in if you would rather the hook land complete.

## Verification

- `tests/modkit/cases/pokemon_level_visible.lua` — the contract through the public mod API: default true with no mod, `false` suppresses, the surface and the mon reach the hook, falling through prints, and a `nil` mon answers rather than throwing.
- `tests/engine/gate_hooks.lua` picks the hook up on its own, because it walks the live catalog rather than a hand-kept list.
- `tests/engine/gate_meta_coverage.lua` is satisfied by the change that introduces the seam, so it never enters the DEBT ledger.

`tests/run_modkit.lua` 33/33. `tests/run_engine.lua` 327/331 — the four failures are the audio-device-reset and hostshell-curl suites, which fail identically on `dev` without this branch (verified by stashing).

## Screenshots

| Before  | After (when enabled via a mod)  |
| :---: | :---: |
| image | image |
| With status (check Pidgey) image | With status (check Pidgey) image |
| image | image |
|  image | image |
| image  |  image |

### DIFF
--- a/docs/modding.md
+++ b/docs/modding.md
@@ -1111,6 +1111,37 @@ own state, preserving selective control outside a battle; a wrapper that only
 owns battle presentation should return `false` only for its active battle or
 text-box state.
 
+`pokemon.level_visible` (RFC 0019) takes a Pokémon's level off the screens
+that print it, without moving anything else on them:
+
+```lua
+mod.hooks:wrap("pokemon.level_visible", function(next, mon, ctx)
+  -- ctx = { where = "battle.enemy" | "battle.player" | "party" | "summary",
+  --         game = <Game> }
+  if myMode.active and ctx.where ~= "party" then return false end
+  return next(mon, ctx)
+end)
+```
+
+It receives `(next, mon, ctx)` and defaults to `true`, so vanilla rendering
+is unchanged. Only an explicit `false` suppresses; `nil` and anything else
+print, so a wrapper that forgets a branch cannot blank a screen by accident.
+The `<LV>` glyph goes with the digits, and on status page 2 so does the
+`<to>` arrow that points at the next level, because an arrow with nothing
+after it is half a sentence. A status condition still replaces the level on
+a battle healthbox exactly as it does in the cart, so hiding a level never
+hides `PSN` or `BRN`.
+
+`ctx.where` names the surface rather than the widget, because the number
+means different things on different screens: on a battle HUD an opponent's
+level is information about them, on the party and status screens your own
+level is information about you. A mode can hide one and keep the other.
+
+**Gen 1 only for now.** The Gen 2 screens keep their own level readouts and
+do not consult this hook, and neither does the Gen 1 PC box list, where the
+level is part of a row label rather than a drawn field. Both are noted in
+RFC 0019 as follow-ups.
+
 `core.logic_speed` receives `(next, game)` once per `Game:logicSpeed()` call
 (once per frame). Vanilla behavior resolves the per-category GAME SPEED
 option (`GameSpeed.CATEGORIES`: overworld/battle/menu) for whichever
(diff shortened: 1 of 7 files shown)


bryanthaboi:
is this a visible change to something that does not exist in vanilla games

campavao:
> is this a visible change to something that does not exist in vanilla games

yeah this would allow for a mod to hide a Pokemon's level in a few different areas (battling, summary, party, etc)

main driving change is with my battle royale mod, I'm capping and auto-leveling mons as time goes on, and I think displaying the level is confusing folks at times, so I just want to hide it visually 

edit: to be clear, this wouldn't affect anything in the base vanilla game itself. Just when a mod is running, it can hide the level display, but it should leave the vanilla game unaffected. I can post some screenshots here in a sec

Issue #1832, opened 2026-08-26
save data checksum inva


asaver:
### One line summary

Error: save data checksum invalid (main data checksum mismatch) when trying to import Pokémon Crystal .sav

### Which game were you playing

Crystal

### Which build are you running

Windows

### Mods off

- [x] I turned off all mods and can still reproduce this

### Version

0.2.26

### Where in the game

Before loading the game

### Screenshot, or a detailed description if you couldn't take one

Error: save data checksum invalid (main data checksum mismatch) when trying to import Pokémon Crystal .sav. This error does not occur with Pokémon Yellow (save loaded successfully)

<img width="1026" height="800" alt="Image" src="https://github.com/user-attachments/assets/497df9ef-d974-4866-84f4-173cba0f69f7" />

### How to make it happen

1- Try to import Pokémon Crystal .sav file

### What should have happened

Game should have loaded a completed save.

bryanthaboi:
save imports for gen 2 not working yet. technically a bug.

asaver:
Oh, my apologies! Thank you for the quick answer, I shall wait patiently.

PR #1844, opened 2026-08-26
Refuse a Gen 2 cart save as a Gen 2 save, not as a broken Gen 1 one

colsonrice:
#1832 reports a Crystal `.sav` failing to import on 0.2.26:

```
save data checksum invalid (main data checksum mismatch)
```

The save is fine. It is being measured with the wrong generation's ruler.

`SaveFileIO.importToSlot` sends anything that is not exactly `SAVE_SIZE` into
`mainChecksumValid`, which is pokered's main-data checksum:

```lua
  if #bytes ~= SAVE_SIZE then
    local check = SaveConvert.mainChecksumValid(bytes)
    ...
    if check == false then
      return false, "save data checksum invalid (main data checksum mismatch)"
    end
```

Gen 2 carts are MBC3+TIMER. A real Gold, Silver or Crystal battery save carries
an RTC footer, so it is 32786 bytes and never 32768. It misses the size test,
falls into the checksum branch, and gets told it is damaged by a rule written
for a different game. Every real Gen 2 cart save takes that path, every time.

The generation check already exists. It just lives too late, inside `importSav`,
behind the size gate that has already rejected the save. This lifts it into
`SaveConvert.importSupported` and `exportSupported` and asks before the bytes
are measured. Both existing guards now route through the same predicates, so the
early caller and the late one cannot describe the same game two different ways.

Nothing changes for Gen 1, and Gen 2 is still refused. It is refused for the
real reason now, in a sentence a player can do something with.

### Tests

`tests/gen2_save_import_message_test.lua` builds a 32786 byte file, the size a
real cart save actually is, and drives it through `importToSlot` for all three
Gen 2 versions. Against the code as it stands it is 20/26, and the failures are
the report itself:

```
FAIL gold: refused AS a Gen 2 save, got: save data checksum invalid (main data checksum mismatch)
FAIL silver: refused AS a Gen 2 save, got: save data checksum invalid (main data checksum mismatch)
FAIL crystal: refused AS a Gen 2 save, got: save data checksum invalid (main data checksum mismatch)
```

It also pins the part that must not move: a genuinely corrupt oversize Gen 1
save is still diagnosed by pokered's checksum.

Registered beside `gen2_save_convert_cli_test` in `tests/run_tests.lua`.

### The codec is deliberately not in this change

`GenSave.lua` is pokered's SRAM window and nothing else, so the guard refusing
Gen 2 is right to be there, and a real Gen 2 codec is the actual answer to
#1832. I am working on one separately. This is only the part where we stop
telling people their save is broken while they wait, and it is the check that
codec will replace rather than work around.

### One existing test changed

`tests/engine/save_import_retry_bug420.lua` replaces `SaveConvert` with a
minimal double, and `importToSlot` now asks it `importSupported` before it
measures the bytes, so the double needed that method. It answers true, which
keeps that case about the thing it actually tests: that `importToSlot` names
the game whose cache to read.

`tests/engine/save_export_portable_bug752.lua` doubles `SaveConvert` too and is
untouched, since `exportActiveSlot` gained no new call.

### DIFF
--- a/src/import/SaveFileIO.lua
+++ b/src/import/SaveFileIO.lua
@@ -78,6 +78,12 @@ function SaveFileIO.importToSlot(source, version, force)
   version = version or GameVersion.get()
   local bytes, readErr = readSource(source)
   if not bytes then return false, readErr end
+  -- The GAME decides before the BYTES do.  Everything below this line judges a
+  -- save by Gen 1's rules -- the size test, and mainChecksumValid, which is
+  -- pokered's checksum -- so a Gen 2 cart save reaching it is measured against
+  -- a rule that cannot match and comes back "checksum invalid" (#1832).
+  local supported, unsupportedWhy = SaveConvert.importSupported(version)
+  if not supported then return false, unsupportedWhy end
   if #bytes ~= SAVE_SIZE then
     local check = SaveConvert.mainChecksumValid(bytes)
     if check == nil then
--- a/src/save_convert/SaveConvert.lua
+++ b/src/save_convert/SaveConvert.lua
@@ -235,6 +235,35 @@ local function gen2CartName(gameVersion)
   return GameVersion.info(gameVersion).displayName
 end
 
+-- Can a cart save for this game cross in or out at all?  Public because the
+-- launcher has to ask about the GAME before it measures the BYTES.
+--
+-- SaveFileIO.importToSlot judges anything that is not exactly SAVE_SIZE with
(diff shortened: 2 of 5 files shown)


bryanthaboi (approved):
verified this against #1832. the gen1 checksum gate is behind the size branch so an exact 32768 crystal save already got the right message, its the ones with the rtc footer that hit the gen1 checksum first. this fixes that path and nothing else. thanks

PR #1850, opened 2026-08-26
Import Gen 2 cart saves

colsonrice:
Gold, Silver and Crystal cart saves import now. Export is still refused,
deliberately, and says so.

Builds on #1844. That one stops the launcher calling a good Gen 2 save corrupt;
this one makes it actually load. If you take #1844 first this rebases cleanly,
and if you would rather have one PR, this supersedes it.

`GenSave.lua` is pokered's SRAM window and nothing else, which is why the guard
refusing Gen 2 was right to be there. This adds `Gen2Save.lua` beside it,
covering pokegold and pokecrystal.

### Nothing is transcribed

`tools/gen2_sram_offsets.py` reads `pokegold.sym` and `pokecrystal.sym` from a
pret build and emits `Gen2Layout.lua`, text table included from
`constants/charmap.asm`. Gen 2 copies a contiguous WRAM block into SRAM bank 1,
so a field's file offset is

```
sPlayerData + (wField - wPlayerData)
```

The generator asserts that relation against `sPokemonData`, which appears on
both sides, rather than assuming it. It also range-guards anything outside
`sGameData..sGameDataEnd`: Crystal's `wPlayerGender` sits before `wPlayerData`
and belongs to `sCrystalData`, and the naive subtraction gives it a plausible
looking offset that means nothing.

Given this file's history with hand-copied numbers, I did not want to add 400
more by hand.

### Gold's table is not Crystal's

They disagree about nearly every field. The same Crystal save, read both ways:

| | Crystal's table | Gold's table |
|---|---|---|
| party count | 6 | **133** |
| money | 500000 | **4196257** |
| play time | 51h | **13113h** |
| species | Typhlosion, Pidgeot, Feraligatr | `[1, 100, 0, 0, 1, 81]` |

All of it with a checksum that validates, which is why they are separate tables
and why the codec refuses a layout mismatch rather than trusting the sum.

### Three things only a real save could catch

* boxes 8 to 14 start at a `0x620` step, not `0x450`, because they cross into
  SRAM bank 3. A uniform stride reports box counts like 243 and 196.
* OTs sit **before** nicknames in `box_struct`. Swapped, a save reads back with
  the player's name in the nickname slot.
* a box mon is 32 bytes with none of the party's computed stats. Reading a
  party-shaped 48 walks into the next mon.

Each of those passes a checksum. So a box's species list is now cross-checked
against the stored Pokemon, which is a free guard on the layout itself.

### The size gate

`mainChecksumValid` takes the game now and asks that generation's rule, so a
real Gen 2 save is no longer measured with pokered's checksum. The oversize
confirmation is also skipped for a Gen 2 cart: the RTC footer is the normal
shape of one, so asking would be a prompt with a single sensible answer, on
every import, forever.

### Tests

`tests/gen2_save_import_test.lua` builds a save with a known Pokemon in the
party and one in a box, reads it back for all three versions, and pins that the
two layouts are not interchangeable. The save is synthesized rather than checked
in, the same rule `save_convert_tests.lua` follows for Gen 1. Set
`POKEPORT_GEN2_SAV_FIXTURE` (and `POKEPORT_GEN2_SAV_VERSION`) to run the audit
at the bottom against your own cart save.

`tests/gen2_save_import_message_test.lua` now covers the size gate end to end: a
32786 byte save imports, without the confirmation.

Three existing tests asserted that Gen 2 import was refused, which is the
behaviour this changes. `gen2_save_convert_cli_test` and `gen2_save_export_test`
now pin that a Gen 2 save reaches the codec and a blank one is turned away by
Gen 2's own check values rather than by version. The bug420 `SaveConvert` double
grows `isGen2Cart` alongside `importSupported`.

`./scripts/test.sh` passes end to end and luacheck is clean.

### What is not here

Export. It is the direction that can damage a save, and it needs the unmodeled
regions reproduced or the real cartridge will not boot what it writes, so I want
it behind a test that boots the output in a real ROM before I send it.

Decode covers the block, the party, all 14 boxes, the four bag pockets, the
pokedex and the event flags. Mail, phone contacts, the unown dex, the hall of
fame and the RTC merge over `Save.newGame` defaults, the same way Gen 1's
importer leaves what it does not model.


bryanthaboi (changes requested):
the sram side of this is really good. i checked all 48 offsets against both sym files and the checksum window, the 16 bit accumulate, low byte first, the crystal $18a gap vs golds zero gap, all correct. thats the part thats usually wrong so nice work.

problem is the save table it builds isnt the shape the engine actually reads. four things:

1. `save.events` is flag id -> bool but the engine wants byte index -> packed byte. scrubEvents does tonumber(true) so every event flag gets dropped. import a post e4 crystal save and you wake up pre elm, no badges, every item ball unclaimed.
2. key items / balls get their own tables but the gen2 bag is one flat save.inventory bucketed by the items pocket. nothing in src reads save.keyItems or save.balls. wTMsHMs is in the layout but never read or written either. so you import and cant cycle or fish or throw a ball.
3. save.position has no map key so it falls thru to save.spawn. import a save in goldenrod and you resume in new bark at goldenrods x/y.
4. mon.status is the raw byte, engine wants "psn"/"brn"/etc. also status = 0 is truthy in lua so healthy mons read as afflicted.

also the importer refuses saves a real cart would load. TryLoadSaveFile falls back to VerifyBackupChecksum and loads the backup copy, this only checks the primary. thats the same false corrupted message #1832 was about. crystals backup is contiguous so its not hard, gold is genuinley harder, fine to do crystal only.

and the new tests are in the t3 tier which dies at line 441 for an unrelated pre existing reason, way before it reaches them. so none of these have ever run, the green headless check isnt evidence about the codec. theyre rom free, move them to run_engine and theyll actually gate.

keep ur explanations in ur PR, way too many explanatory comments in this code keep code comments simple and instead explain rationale in the PR

colsonrice:
Reworked. Every one of these was right, including the one that mattered most.

**The tests had never run.** They were registered in `run_tests.lua`, which dies
at line 22 in `Data.lua` needing ROM-derived tables, so it never reached them.
Not line 441 in my checkout, but the same conclusion, and worse than I thought:
the green headless check was not evidence about the codec at all. They live in
`tests/engine/` now and actually gate.

**The engine is keyed by name, not just for the bag.** Chasing your inventory
point I found it runs through `mon.species`, `mon.moves`, `mon.item` and the
pokedex keys too. `data.pokemon` is indexed by `"TYPHLOSION"`. So the codec now
crosswalks through the generated tables the way `GenSave.crosswalks` does for
Gen 1, off the same `index` field the Gen 2 extractor writes. My PR text claimed
Gen 2 "needs no crosswalk tables"; that was flatly wrong and is gone.

Worth saying plainly: my tests passed the whole time because they asserted the
numbers I had chosen myself. They were confirming my own misreading.

The four shapes:

1. `events` is byte index -> packed byte now, `Save.EVENT_BYTES` of them.
   Confirmed `scrubEvents` runs `tonumber` over the value and `tonumber(true)`
   is nil, so every flag was being dropped.
2. One flat `save.inventory` keyed by item id. `keyItems` and `balls` are gone,
   and the TM/HM pocket lands there too via each item's `tmNumber`.
3. `position.map` carries the map id, resolved from the generated maps by
   (group, number). `Save.lua:923` is `save.position.map or save.spawn`.
4. `mon.status` is an `ItemEffects.STATUS_CLASS` key with `statusTurns` beside
   it, and nil when healthy.

**The backup fallback is in, Crystal only as you suggested.** The generator
emits Crystal's backup layout from `sBackupGameData`, which is the primary's
shape shifted by one bank. Gold and Silver split theirs across
`sBackupPlayerData1/2/3` and get none. Tested by corrupting the primary
checksum on a real Crystal save and watching it still open.

Comments trimmed, rationale moved here.

One thing I did not do: the pokedex and bag still fall back to the raw cart
number when the crosswalk has no row for an id, rather than dropping it. An id
this build cannot name is still the player's, but say the word if you would
rather it were dropped or refused.

PR #1852, opened 2026-08-26
Export Gen 2 cart saves

colsonrice:
Closes the round trip. Gold, Silver and Crystal progress made in this project
writes back to a cartridge save the real game boots.

Builds on #1850, which is the import half. Take that one first and this rebases
cleanly.

### Template-aware, and template-required

Gen 2 SRAM holds a great deal this codec does not model: mail, the phone book,
the Unown dex, the hall of fame, the RTC, the current map cache. The real game
trusts all of it on CONTINUE. Writing zeros there produces a save whose checksum
is perfect and whose cartridge does not boot, which is what Gen 1 shipped twice.

So `encode` writes into the image the save was decoded from, and a save with no
cartridge behind it is refused rather than invented:

```
this save has no cartridge image to write back into, and a Gen 2 save built
from nothing does not boot on real hardware
```

That refusal is the known limitation, and it is deliberate. A game begun inside
this project has no lineage to write back into yet.

### Only the primary copy

The game keeps a backup save and rewrites it from the primary on every
successful load:

```
TryLoadSaveFile:
	call VerifyChecksum
	jr nz, .backup
	...
	call SaveBackupPlayerData
	call SaveBackupPokemonData
	call SaveBackupChecksum
```

So a stale backup heals itself on the first CONTINUE, and export does not need a
second offset mapping to stay consistent. Worth knowing before writing one:
Crystal arranges its backup differently from Gold and Silver.

### Two bugs real data caught in the text encoder

* The cart's table carries the ligature halves `PO` and `KE`, which spell
  `<POKE>` together. A reverse map that accepts them turned an ordinary name
  containing "PO" into `0x70` where the cart had `0x8F`, a plain P.
* `#glyph` counts **bytes** in Lua, so every multi-byte glyph was dropped from
  the reverse map and came back as `?`. Not a corner case: that is NIDORAN and
  every name carrying an accent or a currency sign.

Both were found by requiring a byte-identical round trip against real saves.
Neither would have failed a checksum.

### Tests

An untouched decode then encode is **byte identical** on real cart saves, all
three fixtures, including the two that carry an RTC footer.

And the one Gen 1 never had. Decode a real save, rename the player, change the
money, level the lead Pokemon, encode, then run the result in `pokesilver.gbc`
and `pokecrystal.gbc` through the title and CONTINUE:

```
Silver  : PASS map=4/3   name=TESTER frame=680 responsive=yes
Crystal : PASS map=21/14 name=ZARA   frame=680 responsive=yes
```

Both reach the overworld on the save's own map, with the NEW name in WRAM, and
still respond to input afterwards. A save that merely checksums correctly can
wedge the real game, so this is the gate that decides whether an export codec
ships at all.

Four suites that pinned export being refused now pin what refuses instead: a
save with no cartridge image behind it.

`./scripts/test.sh` passes end to end and luacheck is clean.


bryanthaboi (dismissed):
same as #1850, this is stacked on it so that one needs to land first.

on top of those:

1. encode never writes the bag at all. no inventory, key items, balls, tms, no wCurBox or box names. they just keep whatever came in from the template. import a save, buy a potion, export, its not there.
2. the rtc footer gets dropped. importToSlot truncates to 32768 so rawImport never has it and encode returns exactly 32768. so you export over your emulators sav, it shrinks, clock resets, daily events and bug contest and time of day all reset and the player eats the clock adjustment penalty.
3. putSharedMon skips 0x1C and 0x1D-0x1E so pokerus and caught data survive positionally. swap two party mons and export, slot 1 has the other mons species and moves but the first ones caught level and location.
4. the round trip test exports a save whose rawImport is the same buffer it decoded from, so its basically copying a buffer onto itself. it cant catch any of 1-3. the real sav audit at the bottom is the right idea but its gated on an env var so ci never runs it and it only range checks instead of comparing known values.

one more, rawImport is 32k of binary stored in the serialized table. i measured it, 40194 bytes of lua source reparsed on every save and load forever for every imported slot. the idea of never writing a gen2 save from nothing is right, just put it beside the slot as a sidecar not in the table.

keep ur explanations in ur PR, way too many explanatory comments in this code keep code comments simple and instead explain rationale in the PR

colsonrice:
Reworked, on top of the #1850 rework. All five were right.

1. **The bag is written now**, all four pockets, bucketed by each item's own
   `pocket`, plus `wCurBox` and the box names. It used to keep whatever the
   template carried, so a potion bought in a session never arrived.

   One thing that fell out of this: without the item table every item buckets
   into ITEM, which holds 20, and a real bag is bigger. I made that a refusal
   rather than a short write. Dropping the overflow silently would be worse
   than the bug.

2. **The RTC footer is carried through.** The image is kept at its full length
   rather than truncated to 32768, so an export lands on the emulator `.sav` the
   same size it left, clock intact.

3. **0x1C-0x1E are written.** You were right that they survived positionally:
   reorder the party, export, and slot 1 had the other mon's caught level and
   location.

4. **The round trip changes things first** and reads them back through a fresh
   decode, in each place export has to reach. The old one exported onto the
   buffer it had just decoded, so every region encode did not write matched by
   construction and it could not have caught any of 1-3. The fixture-gated audit
   exports a real cart save too and pins that the image keeps its size.

5. **`rawImport` is out of the table**, into `saves/<version>/<id>.cart` beside
   the slot. Your 40194 number is why; it was reparsed on every save and load
   for every imported slot, forever.

Two more the byte-identical requirement caught in the text encoder, both mine:
the ligature halves `PO` and `KE` were in the reverse map, so a name containing
"PO" wrote `0x70` where the cart had a plain `0x8F`; and `#glyph` counts BYTES
in Lua, so every multi-byte glyph was dropped and came back as `?`, which is
NIDORAN and every accented name.

On only writing the primary: `TryLoadSaveFile` rewrites the backup from the
primary as soon as `VerifyChecksum` passes, so a stale backup heals itself on
the first load. That is also why I did not map Crystal's backup a second time
for export; it arranges it differently from Gold and Silver.

Comments trimmed, rationale here.

PR #1897, opened 2026-08-28
Add additive TILT billboard masking API

ShaneMcGovernIE:
## Summary

- add an optional additive `render_pipelines.drawBillboards(ctx)` stage inside built-in TILT
- expose a read-only semantic collision-map view plus cached map cutout, source-mask, and billboard queue helpers
- integrate source masking and Y-sorted upright cards across Gen 1, Gold, Silver, and Crystal
- preserve nearby terrain when removing the flat source silhouette, preventing duplicate flat/upright structures

This is a presentation-only API. It does not change collision, movement, warps, scripts, saves, battles, or link behavior.

## API

During built-in TILT, an enabled pipeline may implement `drawBillboards(ctx)`. The context provides:

- `ctx.map`: semantic map dimensions and collision-derived `cell(x, y)` flags
- `ctx.drawMapCutout(...)`: draw a cached alpha card for a seeded map component
- `ctx.maskMapCutout(...)`: remove that same component from the active ground canvas
- `ctx.billboard(worldX, worldY, drawFn)`: merge the card into the normal upright-object Y-sort
- `ctx.mapScale`: map-pixel to world-pixel scale

`maskMapCutout` returns `false` when the region cannot be masked safely, allowing mods to skip the corresponding upright card and avoid partial rendering.

## Verification

- `luajit tests/mod_render_tests.lua` — 62/62
- additive billboard/cutout suites — 59/59
- changed Lua modules compile with `luajit -b`
- `luajit tests/run_engine.lua` — 367/368 in the restricted workspace; the sole failure was temporary-fixture write permission
- `luajit tests/engine/required_import_streaming_test.lua` — 11/11 when rerun with repository write access
- `python3 tools/modkit.py docs`
- `git diff --check`

Opened as a draft for API and renderer review.

### DIFF
--- a/docs/mod-api-gen2-compat.md
+++ b/docs/mod-api-gen2-compat.md
@@ -328,6 +328,13 @@ you write against it:
   (`Pipelines.rows` is read only from `src/ui/OptionsMenu.lua`), so a Gold
   player reaches one by its `hotkey`.
 
+  The additive **`drawBillboards` stage is live on Gold, Silver, and Crystal**
+  when the built-in TILT mode is active. It receives the same semantic
+  collision-map view as Gen 1 plus `drawMapCutout`, `maskMapCutout`, and
+  `billboard` helpers. Source silhouettes are masked from the tilted ground
+  before queued cards join the normal upright-object Y-sort, preventing the
+  flat and upright copies from being drawn together.
+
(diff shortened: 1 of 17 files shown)


bryanthaboi:
5 million files changed

PR #1903, opened 2026-08-28
fix(gen2): preserve caller identity for deferred phone calls

AverageConsumer:
## What this fixes

Mom can call after a trainer battle to say she bought something. That call waits until the current overworld script is finished. The shared caller value can still contain an earlier trainer contact, so the caller box may show that stale trainer even though Mom is speaking.

The stale trainer does not need to be the trainer from the immediately preceding battle, and the player does not need to have accepted that trainer's phone number.

## How it works

- Store the intended contact on a queued phone script.
- Restore that contact immediately before the queued script starts.
- Include Mom's contact in her shopping-call descriptor.

This only changes queued phone scripts that carry caller metadata. Ordinary queued scripts and existing direct phone calls keep their current behavior.

## Why this is safe

The metadata is additive. `runQueuedScript` behaves exactly as before unless the queued table contains `phoneContact`. There is no mod API or event-contract change.

## Testing

- Added a regression test that queues Mom's call, simulates a stale trainer caller value, and verifies that the call still starts as Mom: 48/48 checks passed.
- A/B tested on Android: the control build reliably showed the stale trainer name; the fixed build showed Mom.
- Gen 2 suite: 135/137 suites pass locally. The same two remaining failures reproduce on unchanged `dev` and are Windows-only path/shell assumptions (`/tmp` and Unix command output), unrelated to this change.
- Lua/JIT source-limit checks: 452/452 passed.

### DIFF
--- a/src/core/gen2/PhoneRing.lua
+++ b/src/core/gen2/PhoneRing.lua
@@ -100,6 +100,7 @@ function PhoneRing.script(call, name, className)
     })
   end
   local rows = {}
+  rows.phoneContact = call and call.contact
   if call and call.delay then
     rows[#rows + 1] = { op = "pause", frames = call.delay }
   end
--- a/src/world/gen2/World.lua
+++ b/src/world/gen2/World.lua
@@ -5040,7 +5040,11 @@ function World:runQueuedScript()
   if not script or self:busy() then return false end
   self.queuedScript = nil
   self.talkNpc = nil
-  return self.vm and self.vm:start(script) or false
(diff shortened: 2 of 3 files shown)


AverageConsumer:
CI note: the engine-lint failure is already present on the current dev base (a9ce423b) and comes from src/online/TeamPick.lua lines 207 and 209. This PR does not touch that file. The headless suites, including the new regression test, pass. I will rebase once the baseline lint fix lands if needed.

Issue #1691, opened 2026-08-22
Exported .sav not valid in emulator


jimjonjimmy:
### One line summary

The .save file that gen1recomp exports cannot be used in an emulator.

### Which game were you playing

Red

### Which build are you running

Multiple platforms

### Mods off

- [x] I turned off all mods and can still reproduce this

### Version

0.2.19

### Discord username (optional)

jimjonjimmy

### Where in the game

From menu to loading the file.

### Screenshot, or a detailed description if you couldn't take one

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/04396e8e-976a-4045-b1bf-0c5da42a42cf" />
When loading the game using in this case Retroarch using the Gambette core. My character gets loaded into a garbled mess of graphics and the game immediately freezes. I have tried editing the save in PKHex but that did not seem to help even though PKHex says the save is valid.

### How to make it happen

1. copy save to emulator save directory
2. match name to rom name
3. load game in emulator
4. continue game

### What should have happened

The game should load up normally. In this case I should have been in the cinnabar Pokémon center.

bryanthaboi:
Wtf pkhex says valid and it STILL doesn't work for u that's crazy I guess we got more debugging to do

jimjonjimmy:
I can try a few more emulators later. I found out it wasn't working from a Pokemon Red to FireRed save convertor that someone recently released.

bryanthaboi:
Prior art on this is #206, same symptom of a garbled overworld and an instant freeze after loading an exported save. That one is closed, so this is the live tracker now. Worth diffing the current exporter against the notes there.

BelardoA:
I get a similar experience when export my red and yellow saves. The only difference is I'm writing the save back to an official cartridge. 

Edit: spelling

ioniccipher-glitch:
The saves I exported from red, blue, and yellow all seem to work on mGBA, and when converted to .dat files they work on virtual console on 3ds.

Edit: forgot to mention the the saves were exported on the 22nd of August 2026.

Edit: also worked for a red save exported 27th of August 2026

Issue #889, opened 2026-08-06
Pokemon yellow can't export save file right


tthiagolourenzi:
### One line summary

Hello, the save file export is still not working on pokemon yellow, 

### Which game were you playing

Yellow

### Which build are you running

Windows

### Were any mods on

No

### Version

0.1.72

### Discord username (optional)

AASnneNPC

### Where in the game

Vermilion City

### Screenshot, or a detailed description if you couldn't take one

<img width="3120" height="1440" alt="Image" src="https://github.com/user-attachments/assets/3ac2d2aa-5e3e-4533-a6e8-e32fc60add27" />

### How to make it happen

1. export te .sav file from the gen1recomp app
2. import in a folder where the emulator can read it
3. open the game and click "continue"

### What should have happened

the save should open normally

### Anything else

I'm sorry for reporting the same error again after reporting yesterday, please if the bug is really gone and it's a problem on my end ignore this bug report.

bryanthaboi:
No need to apologise, retesting is useful. #838 was the Yellow export fix and it shipped in 0.1.72, which is the version you are on, so this is worth chasing. Two things that would help: confirm the .sav was exported from 0.1.72 itself rather than an older export you already had, and tell me which emulator or flashcart you loaded it in. Keeping this open.

tthiagolourenzi:
Yes, the .sav was exported from the 0.1.72 version of gen1recomp. I tried to export to these emulators: MyOldBoy! Pizza Boy GBC, Retrarch.

unxpected-uxp:
Exporting from 0.1.75 and trying to import in mGBA is pretty much the same. Red version.
<img width="640" height="576" alt="Image" src="https://github.com/user-attachments/assets/b32e54a5-fb98-4ddd-a0b0-45830195fb6b" />

unxpected-uxp:
And in blue. For this test I disabled all mods (0 active) and created a new game. Just in the players room without even moving. Save -> Export -> load in mGBA
<img width="640" height="576" alt="Image" src="https://github.com/user-attachments/assets/9f4f914c-d090-46f4-a646-bfbcfbff843c" />

PR #1918, opened 2026-08-28
Carry the map's object state when a Gen 2 export moves maps

colsonrice:
Finishes what #1852 named as its known limitation. A Gen 2 export writes into
the cartridge image the save came from, and everything that image holds for
ITS map is right only there. Export a save that has moved on since, and the
real game continues onto the new map with the old map's people in it. Anyone
who imports a save, plays on, and exports hits this: the moment the player
walks through a door, Export writes a save the cartridge garbles.

### Why the game cannot save us here

Continuing a Gen 2 save re-derives the map but not its people.
MapSetupScript_Continue runs LoadMapAttributes_SkipObjects: attributes,
blocks, graphics and palettes all come back out of the ROM by
wMapGroup/wMapNumber, but ReadObjectEvents is skipped on purpose, so
wMapObjects, the object structs, the object masks and
wCurMapObjectEventCount/Pointer are trusted straight out of SRAM:

```
MapSetupScript_Continue:
	mapsetup DisableLCD
	mapsetup InitSound
	mapsetup LoadMapAttributes_SkipObjects
	...
```

A cartridge save always has that window because the game wrote it. An export
that changed wMapGroup and wMapNumber under it produced the garbled
overworld: the new map's blocks with the old map's NPCs, object scripts and
event flags walking around in them. Gen 1 had the same class of bug and
closed it with MapContext (#889, finished in #1691); Gen2MapContext is that
module for Gen 2.

### What gets rebuilt, and from where

Only when the export actually moved maps. A same-map export leaves the
template's object window untouched, byte for byte, so #1852's byte-identical
round trip still holds and is still pinned by the tests.

On a move, Gen2MapContext replays ReadObjectEvents exactly: each
object_event's thirteen ROM bytes behind a struct id of -1, the 0 / -1 empty
pattern for the rest, cleared NPC structs (ClearObjectStructs), zeroed
object masks, wObjectFollow reset, and wCurMapObjectEventCount/Pointer set
to the new map's count and its list's ROM address. The player keeps the
template's own struct and map object, standing and idle exactly as an
in-game SAVE leaves them, re-anchored to the new tile.

The extractor already captured every object field this needs, script
pointers included; it now also records the object list's address. The Gen 2
cache tags bump, so a cache from before that field refuses with "re-import
the ROM" instead of exporting wrong, the same contract #1691 set for Gen 1.

### Run in the real games

Same gate #1852 used: a real save, moved by this codec, run in the real ROM
through the title and CONTINUE, probed in WRAM rather than eyeballed.

```
Crystal : 21/14 -> 21/15 CELADON_MANSION_ROOF   landed 21/15 at the exported tile
          roof attributes and blocks loaded from ROM ($2B:$44B4), walks, responds
Silver  : 24/4  -> 24/7  PLAYERS_HOUSE_2F       landed 24/7 at the exported tile
          blockdata bank and pointer are the map's own ($37:$495D)
```

### Known limitation, deliberate

NPCs standing inside the very first screen are not drawn until the player
crosses a screen edge or re-enters the map. The engine spawns struct-id -1
objects from its edge scan (CheckObjectEnteringVisibleRange), which is also
how it heals a real save whose NPC wandered off screen before SAVE, so
everything past the first screen behaves exactly as vanilla. Spawning the
in-view ones too means deriving each sprite's VRAM tile slot
(GetSpriteVTile) at encode time; that is derivable from the map's sprite
set and is the natural follow-up, but it is graphics bookkeeping this
change does not need in order to be correct about the world.

Also unchanged and deliberate: object masks start at zero, so an object a
scene script would hide can show until that scene next runs. Objects hidden
by their own event flag, which is nearly all of them, hide correctly,
because those flags travel with the save.

### Tests

tests/engine/gen2_save_export_map_objects.lua, 84 checks across both
layouts: same-map byte identity, every derived byte of a full and a minimal
object, the empty pattern, the player re-anchor, cleared structs, masks,
follow, count and pointer, the checksum reseal, and both refusals (unknown
map, stale cache). Neighbouring save suites pass untouched and luacheck is
clean on every file changed.

### DIFF
--- a/src/import/CacheContract.lua
+++ b/src/import/CacheContract.lua
@@ -10,7 +10,13 @@ local CacheContract = {}
 
 CacheContract.FORMAT = "rom-cache-v10:"
 CacheContract.VERSION_FORMAT = {
-  crystal = "rom-cache-v10-crystal4:",
+  -- v11: Gen 2 maps carry their object list's ROM address, which a .sav
+  -- export re-anchoring a save onto another map writes back into
+  -- wCurMapObjectEventsPointer. A v10 cache has no address to write, and
+  -- such an export is refused until the ROM re-imports.
+  gold = "rom-cache-v11:",
+  silver = "rom-cache-v11:",
+  crystal = "rom-cache-v11-crystal4:",
 }
 CacheContract.MARKER_PATH = "rom-cache.complete"
 
--- a/src/import/RomExtractorGen2.lua
+++ b/src/import/RomExtractorGen2.lua
@@ -1208,6 +1208,11 @@ function RomExtractorGen2:readMapEvents(bank, address, spriteOrder)
 
   local objectCount = self.rom:byte(bank, cursor)
   cursor = cursor + 1
+  -- The bus address of the first object_event, which is exactly the pointer
+  -- ReadObjectEvents (home/map.asm) leaves in wCurMapObjectEventsPointer: DE
+  -- after the count byte. A .sav export re-anchoring a save onto this map
+  -- writes it back so a later ReloadMapEvents reads the right list.
+  local objectEventsAddr = cursor
   local objects = {}
   for i = 1, objectCount do
     local spriteId = self.rom:byte(bank, cursor)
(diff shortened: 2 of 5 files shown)


castdrian:
this needs a rebase.

colsonrice:
Rebased onto dev. One conflict, the crystal cache tag: dev moved it to crystal4 while this branch carried v11-crystal3, so it is now rom-cache-v11-crystal4 and a stale cache invalidates for both changes at once. The new suite is 84/84 on the rebased base, the neighbouring save suites pass untouched, and luacheck is clean on every file changed.

PR #1973, opened 2026-08-30
fix(checkpoint): keep poisonSteps across a restore (#1971)

### LINKED ISSUE #1971: Checkpoint restore fails 3 steps out of 4: the map-enter path zeroes save.poisonSteps, then verification rejects the mismatch
### What happens

`Checkpoint.restore()` rejects most valid checkpoints:

```
Checkpoint restoration failed: restored state differed at $.save.poisonSteps
```

It reproduces on any checkpoint captured while `save.poisonSteps ~= 0`, which is three steps out of four in ordinary overworld play.

### Why

`poisonSteps` is a plain step counter, not a poisoned-only one — it advances on every step regardless of party status (`src/world/OverworldController.lua`):

```lua
local interval = FieldDefaults.world(Game.data, "poisonStepInterval") or 4
save.poisonSteps = ((save.poisonSteps or 0) + 1) % interval
```

Restore re-enters the map, and the map-enter path zeroes it (correctly, mirroring `ClearVariablesOnEnterMap`):

```lua
-- Game:restoreCheckpointSave
self.stack:push(self.overworld, loaded.player.map, ...)

-- OverworldController.setMap
if not (opts and opts.seamless) then Game.save.poisonSteps = 0 end
```

`Checkpoint.restore` then re-captures the applied state and compares it against the checkpoint. The applied value is `0`, the checkpoint says `1`..`3`, the comparison fails, and the whole restore is rolled back.

So the zeroing is a side effect of reusing the map-enter path during restore, not a decision about checkpoint semantics — but verification treats it as a mismatch.

### Suggested fix

Re-apply the checkpoint's own value after the map push, so the field survives the round trip:

```lua
function Game:restoreCheckpointSave(loaded)
  self.save = loaded
  self:adoptSave(loaded)
  while self.stack:top() do self.stack:pop() end
  local poisonSteps = loaded.poisonSteps
  self.stack:push(self.overworld, loaded.player.map,
                  loaded.player.x, loaded.player.y, loaded.player.facing,
                  { via = "checkpoint", checkpoint = true, freshBoot = true })
  -- setMap zeroes poisonSteps on map entry (ClearVariablesOnEnterMap); a
  -- checkpoint restore is not a map entry from the player's point of view, and
  -- verification compares this field.
  self.save.poisonSteps = poisonSteps
end
```

Alternatively, exclude the field from checkpoint verification — though keeping the value seems truer to what a restore means.

### Impact

Any mod using the public checkpoint API is affected, since nothing about the capture side is unusual: the value is whatever the step counter was at capture. Observed with a save-history mod of my own, and by inspection this affects `Checkpoint.resume()` (title-side) on the same path.

### Reproducing

1. Start a playthrough, walk an odd number of steps, capture a checkpoint.
2. Restore it.
3. Fails unless the step count happened to land on a multiple of 4.

Engine 0.2.39, macOS.

mleo2003:
Fixes #1971.

`Checkpoint.restore()` rejects most valid checkpoints:

```
Checkpoint restoration failed: restored state differed at $.save.poisonSteps
```

`poisonSteps` is a plain step counter — `(poisonSteps + 1) % 4` on **every** step, not only while a mon is poisoned (`OverworldController:applyFieldPoison`) — so it is non-zero three steps out of four in ordinary play.

Installing the restored world re-enters the map, and the map-enter path zeroes the counter, correctly mirroring `ClearVariablesOnEnterMap`. `Checkpoint.restore` then re-captures the applied state and compares it against the checkpoint, so the field it had just discarded fails the comparison and the whole restore rolls back.

A restore is not a map entry from the player's point of view: the counter belongs to the state being restored. This carries it across the push.

### Why it stayed hidden

The two autosave triggers a checkpoint consumer naturally uses — `map.entered` and `player.warped` — are emitted from inside the very paths that zero the counter:

```
OverworldController.lua:377   Game.save.poisonSteps = 0
OverworldController.lua:572   Runtime.emit("map.entered", …)
OverworldController.lua:4664  Game.save.poisonSteps = 0
OverworldController.lua:4783  Runtime.emit("player.warped", …)
```

So those captures hold `0` and restore cleanly. Only a capture taken at an arbitrary step — a manual quicksave, or one tied to the ordinary `SAVE` — carries a non-zero value. The T4 title-checkpoint tier misses it for the same reason: its save is fresh, so the counter is already `0`.

I found it building a mod that captures on `save.writing`, which is the one trigger uncorrelated with map entry.

### A stale comment, corrected

The comment above the push said `Checkpoint.resume` was this method's only caller. Both callers arrive through `Checkpoint.apply` — which serves `resume` from the title session and `restore` from a settled runtime — and that is exactly why the map-entry side effects matter here.

### Tests

`tests/engine/restore_poison_steps_bug1971.lua` covers `1..3` and `0` across a restore, and pins the two behaviours that must **not** change:

- a plain map entry still zeroes the counter
- a seamless connection crossing still carries it across

On `main` it fails `3/6` with `got 0, want 3`; with this change it passes `6/6`.

Also run locally: the six existing checkpoint suites (`battle_checkpoint_restore` 43/43, `battle_checkpoint_capture` 29/29, `battle_checkpoint_boundary` 21/21, `restore_standing_trigger_bug1547` 6/6, `resume_boot_music_no_fade` 13/13, `scripted_battle_checkpoint_test` 20/20) and `scripts/test.sh --quick` — **ALL TIERS PASSED**.

Verified end to end as well: with this patch applied to a 0.2.39 build, checkpoints captured at `poisonSteps` 1–3 restore live, where the same records were refused before.

### DIFF
--- a/src/core/Game.lua
+++ b/src/core/Game.lua
@@ -1428,12 +1428,20 @@ function Game:restoreCheckpointSave(loaded)
   self.save = loaded
   self:adoptSave(loaded)
   while self.stack:top() do self.stack:pop() end
-  -- freshBoot unconditionally: Checkpoint.resume (src/core/Checkpoint.lua)
-  -- is this method's only caller, and it is itself gated to the title
-  -- session (isTitleSession).
+  -- setMap zeroes poisonSteps on every non-seamless map entry, mirroring
+  -- ClearVariablesOnEnterMap.  Re-entering the map is how a restore installs
+  -- the world, but it is not a map entry from the player's point of view, and
+  -- Checkpoint.restore verifies the applied state against the checkpoint --
+  -- so the discarded counter failed the comparison and rolled the whole
+  -- restore back three steps out of four (#1971).
+  local poisonSteps = loaded.poisonSteps
+  -- freshBoot unconditionally: both callers arrive through Checkpoint.apply
+  -- (src/core/Checkpoint.lua), which serves Checkpoint.resume from the title
+  -- session and Checkpoint.restore from a settled runtime.
   self.stack:push(self.overworld, loaded.player.map,
                   loaded.player.x, loaded.player.y, loaded.player.facing,
                   { via = "checkpoint", checkpoint = true, freshBoot = true })
+  self.save.poisonSteps = poisonSteps
 end
 
 -- Install a reconstructed battle without calling BattleState:enter(), whose
--- a/tests/engine/restore_poison_steps_bug1971.lua
+++ b/tests/engine/restore_poison_steps_bug1971.lua
@@ -1,0 +1,80 @@
+-- poisonSteps is a plain step counter -- (poisonSteps + 1) % 4 on EVERY step,
+-- not only while a mon is poisoned (OverworldController:applyFieldPoison) --
+-- so it is non-zero three steps out of four in ordinary play.
+--
+-- Restoring a checkpoint re-enters the map, and the map-enter path zeroes it
+-- (setMap, mirroring ClearVariablesOnEnterMap). Checkpoint.restore then
+-- re-captures the applied state and compares it with the checkpoint, so the
+-- field it just discarded fails the comparison and the whole restore rolls
+-- back: "restored state differed at $.save.poisonSteps" (#1971).
+--
+-- A restore is not a map entry from the player's point of view; the counter
+-- belongs to the state being restored.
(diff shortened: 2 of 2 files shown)


bryanthaboi:
Target dev not main

mleo2003:
After AI giving me all kinds of bad advice... it is now fixed.

PR #2007, opened 2026-08-31
fix(core): cut per-frame allocation churn and GC pressure on the main loop

### LINKED ISSUE #2005: Constant walking stutter on weak ARM handhelds (RK3326) from per-frame allocation churn and forced per-frame GC step — jitter, not CPU load
## One line summary
Constant walking stutter on weak ARM handhelds (RK3326) caused by per-frame allocation churn and a forced GC step every frame on the main loop — not by CPU load.

## Which game were you playing
Red

## Which build are you running
Linux (ARM sbc portmaster build)

## Mods off
[x] I turned off all mods and can still reproduce this

## Version
0.2.41

## Where in the game
Overworld walking, Main thread pegged at 100% of one core,

Measured on an RG351MP (RK3326 @1.5GHz, 1GB RAM, dArkOS), PERFORMANCE tier LOW, mods off:

Three jitter sources verified in sourca, this may be concient choices in the code, if they are please explain them on an answer to this issue so its documented for future devs!:

1. src/core/Game.lua:307 and :392 — ModRuntime.call gets an inline closure, allocating a fresh function 60x/sec
2. src/core/Game.lua:424 — collectgarbage("step", 1) every rendered frame.
3. main.lua:41 — checkEmergencyQuit calls love.joystick.getJoysticks() (fresh table) twice per frame to guard a 5-second hold combo

## How to make it happen
Walk in the overworld on an RK3326-class device (R36S, RG351MP). Constant micro-stutters, independent of fps cap, governor, or mods.

## What should have happened
Hoist the closures to module level (behavior identical), space the GC step (e.g. every 4th frame — explicit frees still do the heavy lifting), cache the joystick list (refresh 1/s; irrelevant vs a 5s hold).

## Anything else
All three fixes tested on the RG351MP: stutter eliminated with mods off AND with 11 mods re-enabled. Average CPU unchanged before/after, which confirms the variance-not-load diagnosis. PR linked below.

This issue is written by AI but corrected by me personally c:

laaledesiempre:
### Summary
The main loop allocates and collects garbage every frame for no behavioral gain. Three per-frame churn/pause sources on the hot path, all platform-independent:

1. `Game:step` and `Game:logicSpeed` pass inline closures to `ModRuntime.call`, allocating a fresh function 60×/sec.
2. `Game:update` forces an incremental `collectgarbage("step", 1)` every rendered frame, so the collector is always working against garbage the frame itself just created.
3. `checkEmergencyQuit` calls `love.joystick.getJoysticks()` (a fresh table) twice per frame to guard a 5-second hold combo.

None of this changes any behavior — it's pure per-frame overhead, and it scales with frame rate, so faster machines pay more of it, not less. On desktop it's invisible; on a weak single-core handheld where the main thread sits at 100% CPU, each millisecond of this jitter drops a frame and shows up as constant walking stutter. That's the key diagnostic detail: average CPU usage is unchanged by the fix, because the cost was variance (pauses), not load.

### Changes
- `src/core/Game.lua`: hoist the `ModRuntime.call` vanilla closures (`input.step`, `core.logic_speed`) to module-level locals. Behavior is identical.
- `src/core/Game.lua`: space the incremental GC step to every 4th rendered frame. The explicit frees the comment refers to still do the heavy lifting; collection stays incremental; the stepping stops competing with the frame budget.
- `main.lua`: cache the joystick list in `checkEmergencyQuit` and refresh it once per second (irrelevant against a 5s hold combo).

### Verification
- RG351MP (RK3326, dArkOS, v0.2.41): eliminates the constant walking stutter, with all mods off and with 11 mods re-enabled.
- Average CPU unchanged before/after (129% → 128% sustained) — consistent with variance-not-load.
- `luajit -bl` syntax check passes.

### Notes
- Closes #2005
- This fix was written by AI, reviewed and tested on real hardware by me c:
- [x] I laaledesiempre checked this code and its something a human would have written


laaledesiempre:
the pr edition is because the original (written by ai) was focusing on the hardware specifications that made us look for this bug originally, instead of talking about the bug, performance improve, code etc. ¯\_(ツ)_/¯

1Jamie:
this one's solid too, and the RG351MP numbers are what matter; stutter gone, average cpu basically the same. that's the variance-not-load story, so the diagnosis lines up.

hoisting those hot-path closures, spreading the gc step, caching joysticks for the force-quit hold; all fair game for what #2005 was about.

one heads-up for the record, not a "please rewrite" thing: the gc step only lives on gen1 Game:update. gen2 never had that every-frame collectgarbage call, so gold wasn't in this bug. fine for red/#2005, just don't want someone assuming it covers both gens since only one actually has that.

but beyond those basic comments (no changes needed) i don't see anything that needs to block merge. thanks for testing it on actual rk3326 hardware and keeping the diff small.

laaledesiempre:
yeah i just tested in G/S/C and seems to not have this issue in the first place, it works almost perfectly. also if there is a need for testing in r36s/like systems, feel free to ask me 4 it!

bryanthaboi (approved):

bryanthaboi (approved):

PR #2039, opened 2026-08-31
perf(core): memoize options decode between saves

### LINKED ISSUE #2038: Skins screen is extremely slow on handheld devices (options file re-read and decoded every frame)
**One line summary**
The launcher's Skins screen runs at ~2 FPS on handheld devices because the options file is re-read and re-decoded from disk on every frame.

**Which game were you playing**
N/A (launcher)

**Which build are you running**
Linux (handheld sbc port; the code path is unchanged on current main)

**Mods off**

(The slow path is vanilla `SaveData.loadOptions` — mods only inflate the size of options.lua, which makes each decode costlier; the per-frame read+decode happens with or without them.)

**Version**
Observed on the sbc handheld port; verified against current main (76a63a4e).

**Where in the game**
Launcher → Skins tab.

**Screenshot, or a detailed description if you couldn't take one**
No screenshot. Open the launcher and switch to the Skins tab: the whole view becomes a slideshow while every other tab runs at full speed.

**Details**
`SaveData.loadOptions()` does a full `getInfo` + `read` + decode of the options file on every call, and the Skins tab calls it once per frame (`buildSkinsPanel` → `RomImporter:_activeSkin()`). SkinStudio's library (`drawLibrary`) has the same pattern. Nothing in the result can change within a frame — options only change through `saveOptions` — so the panel pays a full disk read + decode per frame for an identical answer.

Measured on a handheld device with an on-device probe: with an options.lua grown to ~745KB, a single read+decode takes ~616ms, so the Skins screen renders at ~2 FPS (other screens hold 60). A memoized decode (re-read only after an actual write) takes the same screen back to 60 FPS — PR with the fix incoming.

laaledesiempre:
### Summary
`SaveData.loadOptions()` reads and decodes the entire options file on every call. Immediate-mode UI calls it once per frame — the launcher's Skins tab re-reads it just to learn the active skin, and SkinStudio's library does the same — so the whole file is pulled from the save volume and re-decoded every frame for a result that cannot have changed. Options only change through `saveOptions` (or `loadOptions`' own recovery write), so the decode can be memoized between writes.

### Changes
- `src/core/SaveData.lua`: per-filesystem decode cache behind a revision counter. `saveOptions` bumps the revision after a verified write; `loadOptions` re-reads only when the revision moved. Callers still receive a deep copy per call, never the cached tree — the no-aliasing contract is unchanged — and the cart overlay keeps applying per call because the active cart can change between reads without any write. The recovery path (tmp/bak promotion + heal) is untouched and populates the same cache.
- Engine-only diff. A regression test for the cache contract (second load served without disk, mutation of a returned table does not poison the cache, save→load visibility, cart overlay, recovery path) is available on request.

### Verification
- Tested on a handheld device: the Skins screen went from ~2 FPS to 60 FPS (an on-device probe showed a single ~616ms decode per session instead of one per frame).
- `luajit tests/run_engine.lua` (T1/T2 ROM-free tier): 470/470 suites pass.

This fix was written by AI, reviewed and tested on real hardware by me c:

- [x] I laaledesiempre checked this code and its something a human would have written

### Notes
- Closes #2038


bryanthaboi:
u gotta target dev

laaledesiempre:
sowwy :c now is targeting dev

bryanthaboi (approved):

Issue #274, opened 2026-07-27
Monster Party Sprites coloration


CullBear55:
### One line summary

The Party monster sprites are weirdly colored in the RED palette option

### Which build are you running

Windows

### Version

0.1.28

### Where in the game

Menu

### Screenshot, or a detailed description if you couldn't take one

To add onto what other bug report said about HP colors on RED preset, the purple on the monster sprites' color palette should become more reddish like how it was in GSC for consistency with the overworld palette on the RED preset.

I honestly don't know why the RED preset is called RED instead of GSC on the options menu when it looks more like GSC.

How it is currently -> preferred palette -> Cruddy little MS Paint edit (Imagine HP bars are green)

<img width="769" height="600" alt="Image" src="https://github.com/user-attachments/assets/4bd387b0-88e3-48f0-a9bb-b30b68c8c159" />

<img width="481" height="471" alt="Image" src="https://github.com/user-attachments/assets/ce7b004f-e9df-4249-b14d-5cccfebda3cd" />

<img width="769" height="600" alt="Image" src="https://github.com/user-attachments/assets/574dbb7e-a157-43db-a91c-bbd4de877a2b" />

### How to make it happen

1. Open party

### What should have happened

The palette on the party screen looks weird right now because of its heavy use of purple and blacks for color palette on both the monsters and the health bars. If it's modified to look more like GSC, I think it would look better for the RED preset since that's how it looks in the overworld and battle screen with green HP bars.

bryanthaboi:
Absorbing #272 into this one: the HP bar colours on the RED preset (black when full, purple when low) are the same party screen palette problem and should be fixed together with the icon colours.

PR #2259, opened 2026-09-11
Give mon menu icons the trueColor opt-out

Code-Grub:
## What happens

If you play with a full-colour icon pack, your Pokemon icons turn white with purple blotches. It doesn't happen on every screen, and it often shows up right after changing the COLORS setting, which makes it look random.

It isn't random. The icons only break on screens that declare an SGB palette, and only in ADVANCED.

## Why

Colour on those screens is applied after the frame is drawn. A shader reads each pixel's red channel and swaps in one of four palette colours:

    p.r > 0.83 ? c0 : (p.r > 0.5 ? c1 : (p.r > 0.17 ? c2 : c3))

That works for the game's own art, which is already four shades of grey. Full-colour art is a different matter. A Pikachu's yellow body is nearly all high red, so it clears 0.83 and every pixel of it comes back c0, which is white in every palette. The darker parts fall into c2, and under ADVANCED's MEWMON palette c2 is a strong purple.

White Pokemon with purple edges.

Art that should be left alone is meant to say so with a `trueColor` flag, and the engine then re-blits that rectangle unshaded. Battle pics do this. Trainer portraits do it. Overworld sprites do it.

Menu icons can't. There is no `trueColor` field on the icons record, and `Sprites.iconPath` returns only a path where `Sprites.path` and `Sprites.playerPic` both return `path, trueColor`. There is nowhere to put the flag and nothing to read it, so the art always meets the shader.

## Why I think this is an oversight rather than a decision

The obvious counter-argument: menu icons are OBJ art, and real hardware could never draw one in full colour. You could reasonably say full-colour icons are out of scope.

What changed my mind is `R.sprites`. Those are the overworld walking sprites, described in the schema as a "16x16 grounded walker" with `paletteSource`, `palette` and `paletteId` for OBJ palette assignment. Same hardware class as a menu icon, drawn through an OBJ palette the same way. They carry `trueColor`.

So the engine already allows this departure for one kind of OBJ art but not the other. There is prior art in the wild too, like Unique Menu Icons, which gives every Gen 1 species its own menu icon.

If I have read that wrong and icons are meant to stay quantized, say so and I will close this.

## How I know it is the icon path

Worth saying plainly: I don't have the icon pack. I worked from screenshots two different mod authors sent me, so this is inference from the code rather than a local repro.

One detail pins it down, though. `obpIcon`, the bake that built-in icons go through, can only output `1.0`, `170/255` or `0.0`. Those land in c0, c1 and c3. It can never produce c2. The broken icons are full of c2 purple, so they never went through the bake, which means they came from an `icons.bySpecies` table entry, which means mod-supplied art.

A second screenshot of the same pack working correctly shows a brown Eevee, a red Charmander and a purple Nidorino side by side. No single four-colour palette can do that, so the art really is full colour per species.

This is the same purple `PartyMenu.lua:188` blames for #274. That fix only covered built-in icon classes, which bake to shades 0, 1 and 3 and so can never show it.

## The change

- `Schemas.lua` gets a `trueColor` field on the icons record, Gen 1 and Gen 2 forms.
- `Sprites.iconPath` takes the flag in, puts it on `ctx` so a `pokemon.icon` hook can flag art it substitutes, and returns it beside the path. Callers taking a single return value are unaffected.
- `PartyMenu.drawIcon` skips the bake for flagged art and reports its rectangle.

Two bits I would like a second opinion on:

Skipping the bake matters as much as skipping the shader. `obpIcon` is itself a four-shade remap on the red channel, so it flattens the art before the zone pass ever runs.

The flag has to beat `name` rather than sit beside it. A `pokemon.icon` hook can hand back full-colour art for a species that resolved to a built-in class through `icons.byDex`, so `name` is still set and the entry-shape check from #274 misses it.

## Tests

Three new checks, each watched failing first:

- `mod_graphics_tests`: a built-in class reports no rectangle, a flagged icon reports 16x16 where it drew, and flagged art doesn't take the bake.
- `mod_qol_hooks_tests`: a `pokemon.icon` hook can set `ctx.trueColor`, and an unflagged icon stays unflagged.

Reverting just `src/` drops all three.

Nothing else moved: `run_modkit` 36/36, `run_engine` 558/567, `run_gen2` 144/146. The engine and gen2 failures match a clean `dev` baseline I ran to compare, so none of them come from this.

No vanilla icon record sets the flag, so a vanilla boot behaves exactly as it did.

### DIFF
--- a/src/mods/Schemas.lua
+++ b/src/mods/Schemas.lua
@@ -1819,14 +1819,19 @@ end
 
 R.icons = {
   semantics = "record", target = "icons.bySpecies",
-  value = f.union{ f.str, f.rec{ image = f.path, frames = f.opt(f.int(1)) } },
+  -- trueColor is the same 4-shade opt-out the pokemon and trainers records
+  -- carry: flagged art skips the OBP bake and reports its rect so the SGB
+  -- pass re-blits it unshaded, instead of being bucketed by red channel.
+  value = f.union{ f.str, f.rec{ image = f.path, frames = f.opt(f.int(1)),
+                                 trueColor = f.opt(f.bool) } },
   gen2Value = f.union{
     -- the assignment form: a species id mapped to a sheet name
     f.str,
     -- the sheet form: width/height are the sheet's pixel size, and every
     -- vanilla sheet is a 16x32 two-frame strip
     f.rec{ id = f.opt(f.str), index = f.opt(f.int(0, 255)), image = f.path,
-           width = f.int(1), height = f.int(1), frames = f.int(1) },
+           width = f.int(1), height = f.int(1), frames = f.int(1),
+           trueColor = f.opt(f.bool) },
   },
   gen2Extra = function(id, value)
     if gen2IconIsSheet(id) then
--- a/src/pokemon/Sprites.lua
+++ b/src/pokemon/Sprites.lua
@@ -113,11 +113,17 @@ end
 -- Resolve a party-menu icon image path for `mon`.
 -- vanillaPath is the path PartyMenu already picked from icons.bySpecies /
 -- def.icon / icons.byDex; the hook may replace it.
--- Returns path (possibly nil).
+-- opts.trueColor is the icon record's own flag: an icons.bySpecies table
+-- entry may carry one, the way a pokemon record carries it for a battle pic.
+-- A hook may also set ctx.trueColor, to flag art it substitutes for art that
+-- carried no flag -- the same ctx contract pokemon.sprite has, and the reason
+-- the flag is read back off ctx rather than trusted from opts alone.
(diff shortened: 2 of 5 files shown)


Code-Grub:
Heads up on the empty check list here. I force-pushed 5f81c24 to 4bd4405 to correct wording in the commit message, and nothing else. Both commits point at the same tree, `0627b2c8de1f2ab4251433bbf2dd50413f3d2895`, so the diff is byte for byte what it was before.

The run on the previous commit passed everything, headless suites and luacheck included: https://github.com/bryanthaboi/gen1recomp/actions/runs/34613419813

The new run is sitting at `action_required`, so it needs a maintainer to approve it before checks appear again. Sorry for the extra click.

PR #2271, opened 2026-09-12
Fix white battle panel when menus are mirrored to another display

AverageConsumer:
In Crystal, using Kanto Gear together with Battle Art Voxel Gen2 can produce a large white rectangle over the 3D battle scene when opening the Bag or choosing a Pokémon. Move selection and the rest of the fight can look fine. This change keeps the battle scene visible while those menus are being used on Gear's display.

The mirrored menus remain on the game's state stack so they can still handle input, but their main-screen rendering is hidden through `screen.render_visible`. The Gen 2 renderer currently sees that a menu is on top and draws an additional panel pass, even when every screen above the already-composed battle is hidden. That pass redraws the ordinary battle background over the 3D scene.

Only run that additional pass when a visible screen actually exists above the widescreen base. Visible text and menus keep their existing drawing behavior, and hidden menus retain their input/update ownership. The fix uses the existing visibility hook and has no dependency on either mod.

Validation:

- Windows A/B test with official Recomp 0.2.59, Kanto Gear 3.2.9-test.4 and Battle Art Voxel Gen2 2.1.1. Both variants used the same prepared Crystal save, settings and mods; their game-code difference was this `Game2.lua` change. The tester confirmed the fix in B.
- The added regression cases fail on the original renderer when one or multiple hidden menus cause the base to be redrawn. All 33 screen-visibility checks pass with the fix, including visible overlays above/below hidden menus and native update ownership.
- On the current `dev` base: Gen 2 render-output tests (9/9), widescreen clipping tests (39/39), and render-viewport tests pass.

### DIFF
--- a/src/core/Game2.lua
+++ b/src/core/Game2.lua
@@ -1779,6 +1779,17 @@ function Game2:paintBattleSurround(w, h)
   G.setColor(1, 1, 1, 1)
 end
 
+-- Mirrored menus stay on the input stack. They must not trigger another
+-- panel pass over a completed widescreen scene when none is visible.
+local function hasVisibleOverlay(stack, base)
+  for i = #stack.states, 1, -1 do
+    local screen = stack.states[i]
+    if screen == base then return false end
+    if stack:renderVisible(screen) then return true end
+  end
+  return false
+end
+
 function Game2:drawScene(w, h)
   local G = love.graphics
   -- render.compose reads this after the scene is drawn; the plain overworld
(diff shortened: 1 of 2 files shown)


bryanthaboi:
Only when using battle art?

AverageConsumer:
That’s the combination where it has been reproduced: Kanto Gear + Battle Art Voxel Gen2 in Crystal. The fix itself isn’t specific to Battle Art. The Gen 2 renderer was doing an extra panel draw even when all menus above the widescreen scene were hidden through screen.render_visible. The regression tests reproduce that without either mod.

PR #2300, opened 2026-09-19
fix(game3): gate the warp-cell walkable repair on the metatile behavior

### LINKED ISSUE #2297: FireRed: Leaving the house through a wall
### One line summary

Wall to the left of front door in Red's home can be walked through and leads outside.

N/A

Android

### Mods off

### Which mods (if you first noticed this with any on)

None

### Version

2.63

_No response_

### Where in the game

Red's House 1F Front Door

Image

### How to make it happen

Go 1 space past the door to the left and go down

### What should have happened

The player should have collided with the wall and not go outside.

### Anything else

At least in Pallet town it only happens at reds house.

Nexhas28:
Closes #2297.

## The bug

In `PalletTown_PlayersHouse_1F` the wall tile directly left of the front-door mat could be walked through, and stepping down on it warped the player out of the house into PalletTown.

That cell, `(3,9)`, is `mid 26, coll 1, behavior 0x00 MB_NORMAL`. The map header carries **four** warps — `(5,8)` and `(4,8)` out to PalletTown, `(10,2)` up to 2F, and a fourth dead one at `(3,9)`. pret never lets the player stand on `(3,9)`, so that warp never fires. But `Collision.installWarps` force-marked *every* map-header warp cell walkable, which turned `(3,9)` into `COLL_DOOR` and opened the wall.

## The fix

`src/core/game3/collision.lua`: the forced-walkable repair now fires only when the extracted behavior is unreadable (the tileset attrs stopped before that mid, so `Collision.behavior` returns `nil`) or is a behavior pret would actually warp on.

`Collision.isWarpMetatileBehavior` covers `0x60`–`0x6F` plus `0x71`, matching `pret/pokefirered/src/field_control_avatar.c`:

- arrow warps `0x62`–`0x65` (`TryArrowWarp`)
- directional stair warps `0x6C`–`0x6F` (`IsDirectionalStairWarpMetatileBehavior`)
- everything `IsWarpMetatileBehavior` accepts: `MB_CAVE_DOOR` `0x60`, `MB_LADDER` `0x61`, `MB_FALL_WARP` `0x66`, `MB_REGULAR_WARP` `0x67`, `MB_LAVARIDGE_1F_WARP` `0x68`, `MB_WARP_DOOR` `0x69`, escalators `0x6A`–`0x6B`, `MB_UNION_ROOM_WARP` `0x71`

Warp **indexing** is deliberately unchanged — a dead warp on a wall is still registered, matching pret's `GetWarpEventAtMapPosition`. Only the grid write is gated.

The `nil` sentinel matters: `Collision.behavior` returns `nil` when the layout is missing, when the pair's behavior table is missing, or when the mid is past the extracted attribute table (`src/import/gba/object_interactions_extract.lua` only fills `0 .. attr_bytes/4 - 1`). Those are exactly the cases the original repair existed for, so they keep it.

## Regression coverage

`tests/engine/firered_house_wall_warp_bug2297.lua` — new, ROM-free, 301 checks. It embeds the pret `13×10` `map.bin` grid and the `building` metatile attributes, translates them through the real `ScriptingCollision.fromCell`, binds the real `Versions.WARPS.FR_PLAYERS_HOUSE_1F`, and asserts:

- the wall left of the door stays solid and is not walkable
- the dead `(3,9)` warp is not live, while the other three still are
- the repair still applies for unreadable attrs and for live warp behaviors
- a readable `MB_NORMAL` wall is neither opened nor indexed
- an exhaustive `0x00`–`0xFF` table pins the predicate

It fails 6 checks against the pre-fix file.

## Verification

- `luajit tests/engine/firered_house_wall_warp_bug2297.lua` → 301/301
- `luajit tests/run_engine.lua` → 585/585 suites passed
- Cross-checked the only other impassable warp cells in the registered maps: `SEVII_ONE_ISLAND` mids 663/747 are `0x069 MB_WARP_DOOR`, so the repair still applies there — no regression

`scripts/lint.sh --gate` could not run in this environment: luacheck 1.2.0 crashes under Lua 5.5 with `attempt to assign to const variable 'field_name'`. Pre-existing and unrelated to this change.


Nexhas28:
Automated CI status notice only (thollander/actions-comment-pull-request), so there is no reviewer feedback to accept or decline and no code change is warranted. I confirmed run 35431717345 concluded success on 89c7959 with all jobs green and all six artifacts present and unexpired; one cosmetic issue is that the Android link points at artifact 10580738956, which 404s, while the real Android artifact id is 10580738955.

PR #2333, opened 2026-09-19
Fixed Android asset importer

TheRealSolidusSnake:
Just updates RomImporter so asset importers like A Link to the Past work properly on Android.

The importer now uses Android's native file picker, stages imported files separately from the normal Pokémon ROM flow, and routes them back to the correct importer when the picker closes. 

The fallback file browser also no longer filters out .smc / .sfc files.


castdrian:
ci checks are failing

PR #2346, opened 2026-09-19
Make FireRed's hardcoded text translatable, and let Strings() translations reorder values

thibautbus:
## Summary

Most FireRed text comes from the ROM and already reaches translation mods, but the game3 runtime still drew many messages, menu labels and prompts as Lua literals: battle messages, field move and item-use results, the bag, party, PC, shop and summary menus, the new-game intro, the Pokédex, the Town Map and the options. A translation mod had no way to reach them. On top of that, Game3 never called `Strings.load` after the mod merge, so FireRed sessions kept whatever catalog the launcher had preloaded.

This branch loads the merged catalog on FireRed boots and routes that text through `Strings()`, following the conventions Gen 1 and Gen 2 already use. It also lets a translation number its format directives, so languages that put two values the other way round from English can follow their official wording.

## Changes

One commit per change, to keep the review manageable:

1. **Load the merged Strings catalog on FireRed boots.** `Game3:_loadMods` now calls `Strings.load(self.data)` after the merge, the way `Game` and `Game2` do. Covered by `tests/game3_strings_catalog_test.lua`.
2. **Battle messages.** Every `say`/`push`/`askYesNo` line uses a whole-sentence template (`Strings("%s fainted!", name)`) instead of concatenated fragments, so a translation can word the whole sentence around each value. A translation can put the values in another order with numbered directives (commit 6). Variants that used to append a fragment ("sharply rose!", "a boosted") are now full templates of their own. "Ally"/"Foe" prefixes, stat names, weather and screen labels, and the FIGHT/BAG/POKéMON/RUN labels are translated where they are shown. The GHOST nickname checks compare against `Strings("GHOST")`, so they still match the translated nickname. Code that read the English text back is fixed too: the double-battle send-out (animated and headless intros) no longer parses the trainer out of the translated "sent out" line, the battle UI paces the move-used line by its catalog wording, and `AnimSeq` recognises the miss, confusion, substitute, move-used and fainted lines it paces animations on through their catalog wording (`tests/game3_battle_translated_text_test.lua`).
3. **Field, item and script messages.** Field move texts, item-use results, Pokédex and summary fallbacks, trainer send-out lines, step events, VS Seeker and multichoice labels. Nature names and ability/move descriptions are translated where the summary asks for them. Trainer intros use one template with a name and one without. Covered by `tests/game3_strings_module_tables_test.lua`.
4. **Menus.** Bag, TM Case, Berry Pouch, party, PC, box storage, shop, summary, save and yes/no windows.
5. **Intro, Pokédex, map and options.** The controls guide, Pikachu pages and Oak's speech, naming titles, the title menu, Pokédex headers and labels, Town Map names and dungeon descriptions, option values and hints, the Help System bar, the Hall of Fame and the Trainer Card.

6. **Numbered format directives.** `Strings()` formats with `string.format`, which fills directives in argument order, so "PIKACHU's ATTACK" could not become the official French "ATTAQUE de PIKACHU". A translation can now write `"%2$s de\n%1$s monte!"`, the POSIX printf form, and `Strings()` passes the arguments in that order, keeping flags and widths (`%3$03d`). A translation numbers all of its directives or none, and may only number arguments the source has (it may leave one out, as the official translations sometimes drop a name); otherwise it falls back to English with one warning, like a wrong directive count today. A translation that still cannot format its arguments now falls back to the English source formatted with them, instead of the raw template. Sources never carry numbered directives, so English and existing catalogs take the unchanged `string.format` path. Covered in `tests/mod_strings_tests.lua`.

7. **Summary descriptions found by the ROM's names.** `SummaryData` looked move and ability descriptions up as `MOVE_<name>`, built from the name on screen. A translation mod renames moves in place, so with one enabled nothing matched and the summary and TM Case showed "---". `Pokemon.install` now keeps the ROM's English names (`Pokemon.romMoveName`/`romAbilityName`) and the lookups use them. Covered by `tests/game3_summary_description_test.lua`.
8. **Ability names.** No registry renames abilities, so battle messages and the summary now pass the name through `Strings()`, keyed by the cart's English name. The identifiers the battle code compares are unchanged.
9. **Place names in the map popup and forest preview.** The popup's section names and floor labels come from `MapSectionsExtract`, and the forest preview (`MapPreviewScreen`) prints the ROM's English section name; both now go through `Strings()`. The name and the floor now go through `Strings()`, the floor as a whole label ("3F", "B1F", "ROOFTOP"), because the European carts number floors differently (3F is "2e" in French).
10. **Save window layout.** The save window printed its stat values at pret's fixed x, placed for the English labels, so a wider translated label ("DUREE JEU", "SPIELZEIT", "TIEMPO J.") ran into its value. The value column now starts past the widest label, with the English gap, and stays where pret has it for English. Covered by `tests/game3_save_menu_layout_test.lua`.
11. **The Easy Chat screen's own text.** The profile, questionnaire and battle-start editor drew its title, both instruction lines, the confirmation, the quit and delete-all prompts, YES/NO, "Select a group." and the DEL. ALL / CANCEL / OK footer as literals. The per-screen strings translate when the screen opens; `FOOTER_BTNS` is a module-level table, so its labels are `Strings.source` and the draw loop looks each up (`tests/game3_strings_module_tables_test.lua`). The words the player picks still come from the ROM's Easy Chat groups. The title keeps the cart's centring but no longer starts left of its ribbon: the longest English title already fills 118 of the header's 128 px. The confirmation's second line stays one shared key across the three screen kinds, as the cart shares one row for it (`gText_IsAsShownOkay`).

## Design notes

- **Module-level tables are translated when read, not when built.** A `Strings()` call at module load runs before any catalog exists and freezes the English text. These tables keep their English sources, marked with `Strings.source()` so the string scanner still finds them, and are translated at use. `FieldMoves.TEXT` does this through an `__index` proxy, so existing `FieldMoves.TEXT.X` reads keep working. The same applies to `PcMenu.TEXT_*`, the PC labels, `ItemsData.POCKET_LABEL`, `VsSeeker.TEXT`, the intro tables and the rival fallback dialogs.
- **Identifiers stay English.** Values the engine stores or compares are left untouched and translated only where they are drawn. That covers menu action lists (`"CANCEL"`, `"EXIT"`, `"SWITCH BOX"`, `"NEW GAME"`), battle command kinds, quest-log event keys, `special:` keys, box names in the save and asset paths. Default box names are stored as `"BOX 3"` and shown as `Strings("BOX %d", n)`, while names the player typed are drawn as they are.
- **Some keys carry a context.** `cartLabel` calls `Strings(value, "option." .. key)`, so a mod can translate the SHIFT/SET battle style differently from other uses of those words; the new-game name choices use `intro.nameChoice` (FIRE the name is not FIRE the type). Without a context entry, the lookup falls back to the plain key.
- With no translation mod loaded, `Strings()` returns its source, so English output is unchanged. Every concatenation rewritten into a template was checked to give the same English string; the one visible change is that effect messages now name abilities as the cart prints them (COMPOUNDEYES, LIGHTNINGROD, through `Abilities.name`) instead of from the identifier (COMPOUND EYES), and every literal `Strings()` call passes as many arguments as its template has specifiers.

## Not covered here

- Item, move, species and map names come from the ROM pack. The literals left in the Lua are only fallbacks for when the pack has no entry.
- Height and weight are still computed in feet/inches and pounds. A template can't turn them into metric units.
- The help system and the quest log read their text from extracted packs, with no registry; the naming screen's keyboard layout is fixed.

## Testing

- `tests/game3_strings_catalog_test.lua`, `tests/game3_strings_module_tables_test.lua`, `tests/game3_battle_translated_text_test.lua`, `tests/game3_summary_description_test.lua` and `tests/game3_save_menu_layout_test.lua` fail on `dev` and pass here. `tests/mod_strings_tests.lua` covers the numbered directives, including the fallback cases.
- The game3 test scripts give the same results as `dev` at every commit of the branch.
- `scripts/test.sh --quick` gives the same results as `dev`. The four tiers that already fail there fail here too: T0 ROM builder version routing, T0 ROM manifest generator pin/overrides, T0 Yellow title OBP eye remap, T2 Gen 2 / Crystal suites.

In game on Windows with the French FireRed translation mod, on the 0.2.65 release and then on this branch. The accented letters in the "after" shots come from #2342, which is now on `dev`.

### 0.2.65 release
Screenshot 2026-09-19 224052

Screenshot 2026-09-19 224112

Screenshot 2026-09-19 224303

Screenshot 2026-09-19 224708

**Stat change, values reordered**
Screenshot 2026-09-19 224738

Screenshot 2026-09-19 224837

Screenshot 2026-09-19 224851

Screenshot 2026-09-19 224341

### This branch
Screenshot 2026-09-19 223132

Screenshot 2026-09-19 223140

Screenshot 2026-09-19 223349

Screenshot 2026-09-19 225103

**Stat change, values reordered**
Screenshot 2026-09-19 225122

Screenshot 2026-09-19 225154

Screenshot 2026-09-19 230112

Screenshot 2026-09-19 223441

Species names stay English in these shots: the `pokemon` registry's name patches are reverted when the field starts, which is a separate issue. The summary's NAME/TYPE/OT/IDNo labels and type badges are ROM graphics, not text.


bryanthaboi:
Purposeful draft?

thibautbus:
@bryanthaboi I wanted to check again before setting the PR as ready. It's ready now!

Issue #2314, opened 2026-09-19
FireRed: hidden items are unreachable — no hidden-item runtime or ITEMFINDER (e.g. Viridian Forest hidden Antidote)


Nexhas28:
Transcribed from the Gen 3 / FireRed QA report ("Havok Gen3 Bugs"), reported against the **9/18 build**. Checked against `dev` (`da6fe887`, 2026-09-19) — not fixed.

### One line summary

FireRed has no hidden-item runtime at all, so the hidden Antidote (and Potion) in Viridian Forest can never be picked up and the ITEMFINDER has nothing to find.

### Which game were you playing

N/A — this is Pokémon FireRed (Gen 3 / `game3`). The dropdown has no FireRed option, so it's recorded as N/A.

### Which build are you running

Not recorded in the QA report.

### Mods off

- [x] I turned off all mods and can still reproduce this

### Which mods (if you first noticed this with any on)

None.

### Version

9/18 build (QA report dated 2025-09-18).

### Where in the game

Viridian Forest, at the hidden-item tiles.

### Screenshot, or a detailed description if you couldn't take one

No screenshot — transcribed from the QA report, listed as:

> Viridian Forest hidden Antidote missing

### How to make it happen

1. Walk into Viridian Forest.
2. Press A on the hidden Antidote tile at (28, 57) — or the hidden Potion at (3, 22).
3. Nothing happens; no item is given, and the ITEMFINDER never reports anything anywhere in FireRed.

### What should have happened

In vanilla FireRed, pressing A while facing those tiles runs `HiddenItemScript`, which announces the find and grants the item once, guarded by the per-item flag. The ITEMFINDER sweep should also point them out.

### Anything else

Vanilla data for the concrete case (`data/maps/ViridianForest/map.json`):

```json
{"type": "hidden_item", "x": 3,  "y": 22, "elevation": 3, "item": "ITEM_POTION",   "flag": "FLAG_HIDDEN_ITEM_VIRIDIAN_FOREST_POTION",   "quantity": 1}
{"type": "hidden_item", "x": 28, "y": 57, "elevation": 0, "item": "ITEM_ANTIDOTE", "flag": "FLAG_HIDDEN_ITEM_VIRIDIAN_FOREST_ANTIDOTE", "quantity": 1}
```

[...]

PR #2372, opened 2026-09-21
fix(game3): keep hidden-item flags in the authoritative Space store (#2364)

### LINKED ISSUE #2364: hidden items respawn
### One line summary

Respawn the hidden item by opening menu

FireRed

Android

### Mods off

### Which mods (if you first noticed this with any on)

_No response_

### Version

0.2.67

_No response_

### Where in the game

Earliest is viridian forest 

Image

### How to make it happen

.1 start a new game, get to viridian forest
.2 grab the hidden antidote
.3 open the start menu and close out of it
.4 grab the hidden antidote again

### What should have happened

Hidden items shouldn't respawn

### Anything else

_No response_

Nexhas28:
Fixes #2364

## The bug

In FireRed, pick up a hidden item, open the start menu, close it — the item is back and collectable again, repeatedly.

## Root cause

Hidden-item flags are ordinary save flags, and the engine keeps exactly **one** authoritative table for those: `Space.store`, built by `Space.activate` → `load_sidecar`. Every gameplay script reads and writes that store; `session.flags` is only the serialization *mirror* of it.

Three functions in `src/core/game3/field.lua` resolved their store with:

```lua
local store = session and (session.store or session)   -- field.lua:233, 258, 318
```

An FRLG session has no `.store` field — `save_schema_firered.lua` persists `flags`/`vars` only — so this silently fell through to the session table and read/wrote `session.flags` directly. Those three sites were the hidden-item lookup, the pickup, and the Itemfinder, and they were the only three places in the codebase using that idiom.

The failure chain:

1. `Field.pickUpHiddenItem` → `Flags.setFlag(store, ...)` writes `session.flags[0x3E9] = true`. `Space.store.flags[0x3E9]` is never set.
2. Invisible at first, because `hidden_item_at` reads the *same* wrong table, so the item correctly looks collected.
3. Opening the start menu → `Hud.openStartMenu` → `Space.persistSession()` → `persist_sidecar` (`space.lua:129`) → `session.flags = Flags.serialize(Space.store).flags`.
4. That assignment overwrites `session.flags` with the **stale** `Space.store` snapshot and erases `0x3E9`.
5. The item is available again on the very next check — no map reload needed.

### Reproduced against the real modules

Driving the reporter's exact steps (Viridian Forest Antidote, flag `0x3E9`):

```
after pickup:            session.flags[0x3E9] = true
after pickup:            Space.store [0x3E9]   = false
after menu open/persist: session.flags[0x3E9] = false
immediately after persist, hidden item available again = true
```

## The fix

`src/core/game3/field.lua` only. Two small helpers:

- `flag_store(session)` — returns `Space.store` when it exists, else falls back to `session.store or session` so headless callers (and the existing #2314 suite) keep working.
- `set_hidden_flag(session, flag, value)` — writes to `Space.store` **and** mirrors `session.flags`, i.e. the exact shape `item_use.lua`'s `sys_flag` and `Field.openDottedHoleDoor` already use.

Then the three hidden-item sites use them. The pickup flag now lands in the authoritative store, so `Space.persistSession()` re-serializes it back into `session.flags` instead of dropping it, and all three functions read the same store so they cannot disagree. The existing "already collected?" guard in `pickUpHiddenItem` then also correctly refuses a repeat pickup.

## Test

New `tests/engine/game3_hidden_item_respawn_bug2364.lua`. `tests/run_engine.lua` globs `tests/engine/` via `tests/tier_runner.lua`, so it is picked up automatically — no runner list to edit. It drives the real `field.lua`/`space.lua`: pickup → `Space.persistSession()` (the call `Hud.openStartMenu` makes; the menu itself needs a live graphics context) → asserts the flag survives, the item is gone, a second pickup is refused, and the Itemfinder reports nothing. Plus a re-entry check, a "no over-correction" check that other hidden items on the same map are unaffected, and a fallback check for headless callers with no active store.

Verified it is a genuine regression guard:

```
exit code without fix = 1   (6 FAILURES, incl. "does not reappear after opening and closing the menu")
exit code with fix    = 0   (16/16 checks passed)
```

## Validation

- `luajit tests/engine/game3_hidden_item_respawn_bug2364.lua` — 16/16 passed
- `luajit tests/game3_hidden_items_test.lua` (#2314) — still passes
- `tests/run_engine.lua` — **594/594 suites passed** (new suite included)
- `./scripts/lint.sh --gate` — 0 warnings / 0 errors in 920 files
- `./scripts/test.sh --quick` — **ALL TIERS PASSED**

## Notes

- `Space.persistSession` is also called from the save menu (`src/ui/game3/save_menu.lua`) and `src/runtime.lua`, so this respawn was reachable by saving too. One fix covers all callers.
- No save-format change, so existing saves are unaffected — no migration needed.
- Out of scope: field-move flags (CUT/STRENGTH/FLASH, `field.lua` ~845/~892) write only to `Space.store` without a session mirror, which may be a separate persistence gap. Gen2 hidden items (`src/world/gen2/HiddenItems.lua`, `save.hiddenTaken`) are a different implementation and are untouched.
- No doc changes: nothing under `docs/` describes hidden items or flag-store ownership.


bryanthaboi:
mightve fixed already unintentionally plz review

PR #2381, opened 2026-09-21
Implement DoDeoxysTriangleInteraction (Birth Island Deoxys event + battle)

Nexhas28:
## What this does

Implements the **Birth Island Deoxys event** — the triangular rock puzzle *and* the Deoxys battle — ported faithfully from [pret/pokefirered](https://github.com/pret/pokefirered).

Previously `DoDeoxysTriangleInteraction` didn't exist anywhere in the repo, and `SetDeoxysTrianglePalette` was an empty stub. Now the whole event works.

## The puzzle

New file `src/core/game3/deoxys.lua`, ported from `src/field_specials.c:2319-2456`:

- **`Task_DoDeoxysTriangleInteraction`** — the rock can be pushed to 11 positions. Each position has a step budget; walk too far and the puzzle resets. Reach the end and Deoxys appears.
- **`IncrementBirthIslandRockStepCount`** — counts your steps on Birth Island, hooked into `StepEvents.onStepTaken`.
- **Palette changes** — the rock glows progressively redder as you push it. Palettes are baked into images at import time, so I added a colour-substitution mechanism (`OwSprites.setObjectPalette`) instead.
- **Two new field effects** — `FLDEFF_MOVE_DEOXYS_ROCK` (67) and `FLDEFF_DESTROY_DEOXYS_ROCK` (68): the rock sliding between positions, and the rock shattering into fragments when Deoxys breaks out.
- **`DoDeoxysTriangleInteraction = 0x1AB`** registered in `Std.SPECIAL`, with real handlers for it and `SetDeoxysTrianglePalette`.

## The battle

The script does `seteventmon SPECIES_DEOXYS, 30`, which expands to `CreateEnemyEventMon`. That handler already existed, so the encounter flows through `Encounters._pendingWild` into `StartLegendaryBattle` — **no new code was needed for the battle to start**. I added regression tests that pin the entire chain end-to-end so it can't silently break later.

## Bug fix: Deoxys has its own battle theme

> *"Remember that Deoxys has its own unique battle theme"*

Good catch — this uncovered a real bug. Both battle-BGM sites gated the Deoxys theme on species **386**:

- FRLG's `SPECIES_DEOXYS` is actually **410** (`include/constants/species.h:419`)
- **386 is `SPECIES_VOLBEAT`** (`species.h:395`)

So Deoxys was silently getting the generic wild battle theme, and a Volbeat battle would have wrongly received Deoxys' theme. Verified against the engine's own species table: `410 -> DEOXYS`, `386 -> VOLBEAT`, `150 -> MEWTWO`.

Fixed by extracting `Audio.legendaryBattleSong()` as the single source of truth, keyed by the FRLG internal species id, matching the `switch` in `src/battle_setup.c:349`. Both call sites now route through it, so `MUS_VS_DEOXYS` (339) actually plays.

## Verification

| Check | Result |
|---|---|
| `tests/game3_deoxys_test.lua` | **24/24 groups pass** |
| `tests/game3_special_events_test.lua` | pass |
| `tests/run_engine.lua` | **593/593 suites pass** |
| `./scripts/lint.sh --gate` | **0 warnings / 0 errors in 970 files** |
| Hand-run `game3_*` battle / field-effect / object / step suites | pass |

The new tests cover the 11 rock positions, the step caps and reset behaviour, the `FLAG_SYS_DEOXYS_AWAKENED` short-circuit, both field effects, the palette ramp, the theme table (including that 386/Volbeat does **not** match), and the full `seteventmon -> CreateEnemyEventMon -> StartLegendaryBattle` chain.

## Note on verification limits

`data/generated/gba/` is empty in this worktree (no ROM imported), so this is verified by logic tests with stubs rather than by eye or ear — I could not visually confirm the rock animation or listen to the theme. The colour substitution and field-effect drawing degrade gracefully when a sheet is missing.


Nexhas28:
Closing — this PR was opened without being requested. The branch is kept for local testing. Sorry for the noise.

